PR #44, opened 2015-11-30
Start set up to use handsontable for ncube tab.

tympollack:
Duplicated NCubeController just in case and not to break anything while I'm messing with things.


tympollack:
@jdereg - lastest commit with handson ncube table has been made.

PR #52, opened 2015-12-23
Introduced promote revision capability.

tympollack:
I wanted a break from HOT.

### DIFF
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -316,6 +316,7 @@
             </div>
             <div class="modal-footer">
                 <button id="compareRevs" class="btn btn-primary btn-sm pull-left">Compare</button>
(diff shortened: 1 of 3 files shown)


jdereg:
I am not merging this pull request as the promotion to tip revision is being done by having the client send the entire JSON string to the server.  Instead, this should be done with a new NCubeController call that is told to promote the version with the appropriate ID to HEAD. 

call("ncubeController.promote", [appId, cubeId])

This way, the entire JSON of the entire is not sent from the client to the server.  Instead a tiny little ajax call is made telling the server what to do.  The server call (promote) loads the cube with the respective ID, and then saves it.  This is an important distinction.  The connection between the client and the server should always be treated like a scarce resource and we should avoid sending large amounts of data when we can.  In this case, we can avoid it.

PR #62, opened 2015-12-31
Axis Ordering

jsnyder4:


src/main/webapp/js/ntwobe.js
```diff
@@ -413,33 +413,46 @@ var NCubeEditor2 = (function ($)
...
+            var i, len;
+            if (localStorage.hasOwnProperty(getStorageKey(AXIS_ORDER)))
+            {
+                var order = JSON.parse(localStorage[getStorageKey(AXIS_ORDER)]);
+                for (i = 0, len = order.length; i < len; i++)
+                {
+                    var axis = cubeAxes[order[i]];
+                    getColumnLength(axis);
+                    axes.push(axis);
+                }
```
jsnyder4: Tribal getColumnLength() knowledge here

tympollack: What purpose is calling it here? At the very least columnLength will get set on the sort function below so I'm not sure you need to call it here if you're not using the value.

jsnyder4: When it wasn't called here, the cells would not display on the screen properly if you re-ordered the axes and had hidden columns. The reason is, the code below (in the else) doesn't get called if you re-ordered the axes. Therefore the column length wasn't getting set within the global axes variable. The new code in this commit doesn't necessarily use column length, but the other handson table code does.

jdereg: This is a case of a 'getter' with a side-effect.  This getter is lazy-initializing the length.  That does not make the getter wrong, as it is common-practice to lazy-init within a getter.

What it does, though, is potentially fool the reader by making code (on quick scan) look read-only (no side affects) yet the code actually is making changes.

PR #70, opened 2016-01-05
First pass at tabbed navigation.

tympollack:
Feel free to reject; this at least gets us set up with tabs and have the ability to open and close tabs.  Still a lot to work on, but I like smaller commits, especially with the team making frequent changes or shifting work between tasks.


src/main/webapp/js/index.js
```diff
@@ -141,6 +170,77 @@ var NCE = (function ($)
         }
     }
 
+    function selectTab(cubeName) {
+        deselectTab();
+        var list = _openTabsPanel.find('ul').children();
```
jdereg: _openTabsPanel.find('ul') appears 3 or more times.  I recommend placing it in a member var.

jdereg:
I am a fan of small frequent commits as well, especially when they work well and the system keeps running nicely.

src/main/webapp/js/index.js
tympollack: Checked in.

PR #76, opened 2016-01-07
Tabs - many changes.

tympollack:


src/main/webapp/js/index.js
```diff
@@ -126,34 +126,35 @@ var NCE = (function ($)
...
+        _openCubes.push(cubeInfo.join(TAB_SEPARATOR));
         localStorage[OPEN_CUBES] = JSON.stringify(_openCubes);
         addTab(cubeInfo);
     }
 
-    function removeTab(cubeName) {
-        cubeName = getProperCubeName(cubeName);
-        delete _openCubes[cubeName];
+    function removeTab(cubeInfo) {
+        var cia = cubeInfo.join(TAB_SEPARATOR);
```
jdereg: What is cia?  Cube Info a?

src/main/webapp/js/index.js
```diff
@@ -126,34 +126,35 @@ var NCE = (function ($)
...
+        _openCubes.splice(_openCubes.indexOf(cia), 1);
         localStorage[OPEN_CUBES] = JSON.stringify(_openCubes);
-        _openTabList.children().filter(function() { return $.text([this]).indexOf(cubeName) > -1; }).remove();
-        if (_selectedCubeName === cubeName) {
-            var nextCubeName = _openTabsPanel.find('li').first().find('a').first().text();
-            _selectedCubeName = nextCubeName.toLowerCase();
-            selectTab(nextCubeName);
+
+        var cubeName = cubeInfo[3];
+        var activeTab = cubeInfo[4];
```
jdereg: Instead of using array positions, e.g. '3', '4', use a Map with well named keys.  { cubeName: '', activeTab: ' '}.  Then make those keys constants.

src/main/webapp/js/index.js
```diff
@@ -177,71 +178,178 @@ var NCE = (function ($)
...
+            tab = _openTabList.children().first();
         }
         tab.addClass('active');
-        localStorage[SELECTED_CUBE] = cubeName;
+
+        _selectedApp = cubeInfo[0];
+        _selectedVersion = cubeInfo[1];
+        _selectedBranch = cubeInfo[2];
+        _selectedCubeName = cubeInfo[3];
+        _activeTab = cubeInfo[4];
```
jdereg: Here as well - please use an object with named keys, instead of positionally identifying the values.  OR at least use constants instead.

src/main/webapp/js/index.js
```diff
@@ -177,71 +178,178 @@ var NCE = (function ($)
...
+        return '<img src="' + imgSrc + '" height="16px" width="16px"/> &nbsp;';
+    }
+
     function addTab(cubeInfo) {
         deselectTab();
-        var cubeName = cubeInfo[SELECTED_CUBE];
+        var imgSrc;
+        for (var x = 0, xLen = _menuOptions.length; x < xLen; x++) {
+            var opt = _menuOptions[x];
+            if (opt.pageId === cubeInfo[4]) {
```
jdereg: constant instead of 4

src/main/webapp/js/index.js
```diff
@@ -177,71 +178,178 @@ var NCE = (function ($)
...
+                imgSrc = opt.imgSrc;
+                break;
+            }
+        }
         var link = $('<a/>')
             .attr('href','#')
-            .addClass('dropdown-toggle')
-            .html(cubeName + '<span class="big-caret"></span>');
+            .addClass('dropdown-toggle ncube-tab-top-level')
+            .html(getTabImage(imgSrc) + cubeInfo[3] + '<span class="big-caret"></span>');
```
jdereg: magic # 3

src/main/webapp/js/index.js
tympollack: I thought about an array with 0-4 being constants for APP_IDX, etc.  But there was a reason I didn’t bc I needed the keys at one point.  Not sure if that’s still relevant.  I can work on changing it.

src/main/webapp/js/index.js
tympollack: Wasting your time.

src/main/webapp/js/index.js
```diff
@@ -177,71 +178,178 @@ var NCE = (function ($)
...
+                            loadStatusListView();
+                            loadVersionListView();
+                        }
+                        showActiveBranch();
+                        loadNCubes();
+                        loadNCubeListView();
+                    }
+                    selectTab(cubeInfo);
+                    loadCube();
+                }
```
jdereg: I like this new way of determining what to to load - makes it do the minimal amount of work based on what changed.

src/main/webapp/js/index.js
```diff
@@ -275,18 +400,16 @@ var NCE = (function ($)
         $('#appTitle')[0].innerHTML = menu['~Title'];
         var ul = $('#menuList');
         ul.empty();
+        _menuOptions = [];
```
jdereg: Do we need to make a change to sys.menu?  If so, we will need to coordinate the promotion of this change with a promotion of a cube in the NCE.

tympollack: No, it functions exactly the same.

src/main/webapp/js/index.js
tympollack: Checked in.

PR #79, opened 2016-01-10
[RA-23651] Bound autosizing on vertical and horizontal

jsnyder4:
Column widths should autosize.
Row heights are not autosized. The height is determined by the new lines in the cell.

### DIFF
--- a/src/main/webapp/html/ntwobe.html
+++ b/src/main/webapp/html/ntwobe.html
@@ -47,6 +47,12 @@
                 </div>
             </div>
         </div>
+        <table class="handsontable" style="visibility: hidden">
+            <tr>
+                <td id="test-cell" class="cell"></td>
+                <td id="test-code" class="cell code"></td>
+            </tr>
+        </table>
         <div id="hot-container"></div>
     </div>
 </div>
--- a/src/main/webapp/js/ntwobe.js
+++ b/src/main/webapp/js/ntwobe.js
@@ -19,6 +19,7 @@ var NCubeEditor2 = (function ($)
     var cubeMap = null;
     var cubeMapRegex = null;
     var axisColumnMap = null;
+    var _columnWidths = [];
     var _cellId = null;
     var _tableCellId = null;
     var _columnList = null;
@@ -743,7 +744,7 @@ var NCubeEditor2 = (function ($)
             }
         };
 
-        var getHiddenColumns = function()
+        var hideColumns = function()
         {
             _hiddenColumns = {};
             var storageKey = getStorageKey(HIDDEN_COLUMNS);
@@ -751,9 +752,6 @@ var NCubeEditor2 = (function ($)
             {
                 _hiddenColumns = JSON.parse(localStorage[storageKey]);
             }
-        };
-
-        var hideColumns = function() {
             for (var i = 0, axisLength=axes.length; i < axisLength; i++)
             {
                 var axis = axes[i];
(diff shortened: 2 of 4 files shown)


src/main/webapp/html/ntwobe.html
```diff
@@ -47,6 +47,12 @@
                 </div>
             </div>
         </div>
+        <table class="handsontable" style="visibility: hidden">
+            <tr>
+                <td id="test-cell" class="cell"></td>
+                <td id="test-code" class="cell code"></td>
+            </tr>
+        </table>
```
jdereg: I see this is where the content is rendered and then measured.  Cool.

src/main/webapp/js/ntwobe.js
```diff
@@ -779,10 +777,99 @@ var NCubeEditor2 = (function ($)
...
         setUpAxisColumnMap();
+        setUpColumnWidths();
+    };
+
+    var setUpColumnWidths = function()
+    {
+        var calcDomWidth = function (value, modifier, type)
+        {
+            var testElement = type == 'exp' ? testCode: testCell;
+            testElement[0].textContent = value;
```
jdereg: I would verify if .textContent is the right thing to use here, as opposed to .innerHtml.  Create a Groovy expression with something like 'println 'There is <html> in <b>this</b> string'.  If it displays correctly, leave it along.  If it has issues, it likely needs to use .innerHtml.

jsnyder4:
I guess I should do my inline comments after the pull request. I thought I did, but maybe I did them on my branch so they don't show on the file diff here.

src/main/webapp/js/ntwobe.js
```diff
@@ -779,10 +777,99 @@ var NCubeEditor2 = (function ($)
...
+                    var columnId, columnName;
+                    var axisName = axes[hotCol].name;
+                    var buttonWidth = calcDomWidth(axisName, 45, null);
+                    var oldWidth = findWidth(0, buttonWidth);
+
+                    var axisColumns = axisColumnMap[axisName];
+                    for (var axisCol = 0, axisColLength = axisColumns.length; axisCol < axisColLength; axisCol++)
+                    {
+                        columnId = axisColumns[axisCol];
+                        columnName = axes[hotCol].columns[columnId].value;
```
jdereg: columns can be URLs or Values.  Is this code handling that?

src/main/webapp/js/ntwobe.js
```diff
@@ -779,10 +777,99 @@ var NCubeEditor2 = (function ($)
...
+                    var columnHeaderId = getColumnHeaderId(hotCol);
+                    _columnWidths.push(topWidths[columnHeaderId]);
+                }
+            }
+        };
+
+        _columnWidths = [];
+        var topWidths = {};
+        var testCell = $('#test-cell');
+        var testCode = $('#test-code');
```
jdereg: These variables are set here, but I can't tell what there scope is from the review.  My main pointi s...are they used?

src/main/webapp/js/ntwobe.js
```diff
@@ -1754,6 +1837,7 @@ var NCubeEditor2 = (function ($)
         data.cells[_tableCellId] = {value:cellInfo.value};
         _cellId = null;
         destroyEditor();
+        reload();
```
tympollack: You should add reload to cell clear as well if you want it to recalculate.

PR #83, opened 2016-01-11
HOT coordinate bar button styles to match search

tympollack:


src/main/webapp/js/ntwobe.js
```diff
@@ -160,8 +160,8 @@ var NCubeEditor2 = (function ($)
         var search = $('#search-container');
         var searchWidth = search.width();
         var coordWidth = windowWidth - searchWidth;
-        $(getDomCoordinateBar()).width(coordWidth - 45); // coord bar text
-        $('#coordinate-bar-move-right').css({left: coordWidth - 20}); // keep the right button to the end
+        $(getDomCoordinateBar()).width(coordWidth - 50); // coord bar text
+        $('#coordinate-bar-move-right').css({left: coordWidth - 24}); // keep the right button to the end
```
jdereg: Can we get constants for the 50 and 24?

src/main/webapp/js/ntwobe.js
```diff
@@ -1414,31 +1410,14 @@ var NCubeEditor2 = (function ($)
...
-        leftButton.addEventListener('click', function() {
+        $(leftButton).click(function() {
             coordBar.scrollLeft = coordBar.scrollLeft - 40;
+            $(this).blur();
         });
 
-        rightButton.addEventListener('click', function() {
+        $(rightButton).click(function() {
             coordBar.scrollLeft = coordBar.scrollLeft + 40;
+            $(this).blur();
```
jdereg: can we get constants for the 40?

src/main/webapp/js/ntwobe.js
tympollack: Just changed to calculated width.

PR #86, opened 2016-01-12
[RA-23651] Bound autosizing on vertical and horizontal

jsnyder4:
fix single axis n-cubes


src/main/webapp/js/ntwobe.js
```diff
@@ -847,21 +845,8 @@ var NCubeEditor2 = (function ($)
             {
                 if (hotCol < axes.length - 1)
```
jdereg: should this 'if' be using colOffset?

PR #87, opened 2016-01-12
Add rule axis label to hide columns modal

jsnyder4:


src/main/webapp/js/ntwobe.js
```diff
@@ -2323,14 +2323,20 @@ var NCubeEditor2 = (function ($)
             var listItem = $('<li/>').prop({class: "list-group-item skinny-lr no-margins"});
             var rowDiv = $('<div/>').prop({class: "container-fluid"});
             var rowLabel = $('<label/>').prop({class: "checkbox no-margins col-xs-10"});
+            var ruleLabel = '';
+            if (item.metaProps !=null)
+            {
+                ruleLabel = item.metaProps.name + ': ';
```
jdereg: Need to validate that metaProps is valid and metaProps.hasOwnProperty('name') here:
if (item.metaProps && item.metaProps.hasOwnProperty('name')) ...

PR #95, opened 2016-01-13
Tabs - now with dropdown menu on overflow

tympollack:


src/main/webapp/js/constants.js
```diff
@@ -36,6 +36,8 @@ var CLASS_HANDSON_CELL_ODD_ROW = ' oddRow';
 var CLASS_HANDSON_CELL_URL = ' url';
 var CLASS_HANDSON_SEARCH_RESULT = ' searchResult';
 
+var TAB_WIDTH = 225;
```
jdereg: I see a 223 above - is that supposed to match this (225)?

tympollack: No, because the tabs have a margin of 2px between, so this accounts for that as well (it's why I made the width 223)

src/main/webapp/js/index.js
```diff
@@ -432,15 +435,80 @@ var NCE = (function ($)
...
+                                _selectedCubeName = cubeInfo[CUBE_INFO.CUBE];
+                                _activeTab = cubeInfo[CUBE_INFO.TAB];
+                                buildTabs();
+                            })
+                    )
+                );
+            })();
         }
+
+        var button = $('#tab-overflow').find('button');
```
jdereg: If #tab-overflow is static in the page, I would extract these searches into member varirable.

PR #105, opened 2016-01-15
Add compare, rollback, commit on tab level.

tympollack:


src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
```diff
@@ -1153,6 +1153,28 @@ class NCubeController extends BaseController
         }
     }
 
+    Object commitCube(ApplicationID appId, String cubeName)
+    {
+        try
+        {
+            appId = addTenant(appId)
+            isAllowed(appId, null, Delta.Type.UPDATE)
```
jdereg: the cubeName should be the 2nd parameter in the isAllowed() call.

PR #106, opened 2016-01-15
Move single-cube functions to tab menu.

tympollack:


src/main/webapp/js/index.js
```diff
@@ -407,7 +406,36 @@ var NCE = (function ($)
                 .append(
                 $('<a/>')
                     .attr('href','#')
-                    .html('Commit Cube')
+                    .html('Show Revision History')
```
jdereg: Can you change the menu text to read 'Revision History...' now that it does more than just 'show' it?  I should have renamed it a long time ago, once it allowed compare, etc.

PR #119, opened 2016-01-21
Change app/ver/status/branch if cube changed as a result of removing tab

tympollack:
…a tab.


src/main/webapp/js/index.js
```diff
@@ -180,8 +180,27 @@ var NCE = (function ($)
...
+                            loadAppListView();
+                        }
+                        loadVersions();
+                        loadStatusListView();
+                        loadVersionListView();
+                    }
+                    showActiveBranch();
+                    loadNCubes();
+                    loadNCubeListView();
+                }
```
jdereg: If we already have this code elsewhere, probably should package it up into a function and re-use it.  I recall similar code.

PR #146, opened 2016-02-15
Set up front-end frameworks; ref ax changes.

tympollack:
See commit comments.


karma.shared.conf.js
```diff
@@ -0,0 +1,90 @@
+var testStyles = [
+    //'src/main/webapp/css/*.css'
+];
+var thirdPartyLibs = [
+    'bower_components/jquery/dist/jquery.min.js',
+    'bower_components/jquery.gritter/js/jquery.gritter.min.js',
```
jdereg: Is this still referencing the massively customized version of Gritter I created for NCE?  We cannot use the standard gritter as it is very old, based on graphics, and basically is no good.  I completely re-wrote it.

pom.xml
```diff
@@ -183,6 +195,123 @@
...
+                <version>${version.plugin.frontend.maven.plugin}</version>
+                <executions>
+
+                    <execution>
+                        <id>install node and npm</id>
+                        <goals>
+                            <goal>install-node-and-npm</goal>
+                        </goals>
+                        <configuration>
+                            <nodeVersion>v0.12.7</nodeVersion>
```
jdereg: hoist these versions out like the rest...?

PR #148, opened 2016-02-18
Add Filtering for Modals

tympollack:
Not sure how that extra if got there.

### DIFF
--- a/src/main/webapp/js/ntwobe.js
+++ b/src/main/webapp/js/ntwobe.js
@@ -13,8 +13,16 @@
     };
 })(jQuery);
 
-var NCubeEditor2 = (function ($)
-{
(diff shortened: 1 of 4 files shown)


src/main/webapp/js/ntwobe.js
```diff
@@ -242,6 +251,32 @@ var NCubeEditor2 = (function ($)
         setUtilityBarDisplay();
     };
 
+    var addModalFilters = function() {
```
jdereg: Is there a way to consolidate this code to index.js, and then have it available to the views via an nce.* API?

tympollack: I removed it from the ntwobe js anyway; but no that really doesn't work because it's a separate document dom.

PR #161, opened 2016-02-25
Introduce visualizer into NCE to get everyone on the same code base s…

tympollack:
…o we can refactor what needs to be extracted as CUBE-specific.

PR #170, opened 2016-03-08
[RA-26932] Refactor items in previous code review PR #161

jsnyder4:


src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
```diff
@@ -303,7 +303,7 @@ class NCubeController extends BaseController
             Visualizer vis = new Visualizer()
             vis.input = [options:options]
             vis.ncube = nCubeService.getCube(appId, cubeName)
-            return vis.run()
+            return vis.run() as Map
```
jdereg: We should add a method to Visualizer called 'buildGraph', 'getGraph' or something like that, and remove the run method.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -129,39 +120,39 @@ class Visualizer extends NCubeGroovyExpression
...
 
-		return [startCube: startCube.name, networkOptions:getNetworkOptions(), groups: groups, levels: levels, scope:scope, nodes: nodes.toArray(), edges: edges.toArray(), message: message]
+		return [startCube: startCube.name, groups: groups, levels: levels, scope:scope, nodes: nodes.toArray(), edges: edges.toArray(), message: message]
 	}
 
-	private void processCube(Map stack, Set visited, List nodes, List edges, Map groups, Map levels, List messages)
+	private void processCube()
 	{
-		Map stackInfo= (Map) stack.values().toArray()[0]
+		stackInfo = (Map) stack.values().toArray()[0]
```
jdereg: This code is poor performing here.  Calling stack.values() causes a Collection to be build of the values side of a Map (O(n)).  Then, it calls .toArray() which is another O(n).  Then it fetches the first element of the array.  This has to be corrected as it could dramatically slow performance. All that should be done is to have the first element of the Stack popped off.  This is why a Map should not be used for a Stack.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -178,26 +169,26 @@ class Visualizer extends NCubeGroovyExpression
...
 
 		String busType = getFormattedBusType(targetTraitMaps)
 
-		Integer targetLevel = stackInfo[TARGET_LEVEL] as Integer
-		addToEdges(targetCube, sourceCube, targetScope, sourceScope, targetTraitMaps, sourceTraitMaps, sourceFieldName, edges, stackInfo[STACK_KEY] as Integer, targetLevel)
+		long targetLevel = stackInfo[TARGET_LEVEL] as long
+		addToEdges(targetCube, sourceCube, targetScope, sourceScope, targetTraitMaps, sourceTraitMaps, sourceFieldName, stackInfo[STACK_KEY] as long, targetLevel)
 
-		if (hasVisited(visited, targetCube.sha1() + targetScope.toString()))
+		if (hasVisited(targetCube.sha1() + targetScope.toString()))
```
jdereg: Should not be calling targetCube.sha1() if we don't have to.  This is an expensive call (the SHA-1 is cached, so not too bad if it is used more than once).  But is there not a better way to track the cubes visited?

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -283,31 +274,31 @@ class Visualizer extends NCubeGroovyExpression
 				}
 				catch (Exception e)
 				{
-					throw new Exception('Exception caught while loading and processing the cube for enum field ' + targetFieldName + ' in enum ' + targetCube.name + '. ' + e.message, e)
+					throw new IllegalStateException('Exception caught while loading and processing the cube for enum field ' + targetFieldName + ' in enum ' + targetCube.name + '. ' + e.message, e)
```
jdereg: I would remove the + e.message part of the string but leave the ", e".  Otherwise the message will be shown twice.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -463,7 +453,7 @@ class Visualizer extends NCubeGroovyExpression
 
 	private static String getTitleClassTraits(Map traitMaps)
 	{
-		StringBuffer traits = new StringBuffer()
+		StringBuilder traits = new StringBuilder()
```
jdereg: traits local variable not used?  Then why have it?

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -497,19 +487,17 @@ class Visualizer extends NCubeGroovyExpression
 
 	private static Map getFields(Map traitMaps)
 	{
-		Map fieldInfo = [:]
+		Map fieldInfo = [:] as TreeMap
 		traitMaps.each{ fieldName, traits ->
-			String exists = traits[R_EXISTS] as String
-			if (!CLASS_TRAITS.equals(fieldName) && exists && exists.toBoolean())
+			Boolean exists = traits[R_EXISTS] as String
+			if (!CLASS_TRAITS.equals(fieldName) && exists)
```
jdereg: Do we still need the check for CLASS_TRAITS coming back from traitMaps ?

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -822,7 +646,7 @@ class Visualizer extends NCubeGroovyExpression
 	private Set getRequiredScopeAllReferenced(NCube startCube) {
 		Set requiredScope = getRequiredScope(startCube)
 		startCube.getReferencedCubeNames().each {
-			NCube refCube = NCubeManager.getCube(applicationID, it)
+			NCube refCube = getCube(it)
```
jdereg: This whole method should not be needed.  The required scope is declared on the 'entry' cube, and it is expect for that cube to contain all of the required scope, including the required scope for it's supporting actors.

jdereg:
This is a start.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
bheekin: We should be able to use targetCube.name instead of targetCube.sha1. The name of the cube combined with the scope should uniquely identify that "this" class instance has already been visited.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -178,26 +169,26 @@ class Visualizer extends NCubeGroovyExpression
 			stackInfo[TARGET_TRAIT_MAPS] = targetTraitMaps
 		}
 
-		if (!targetCube.name.startsWith(RPM_CLASS)) { //TODO: Can be removed if no exceptions occur
-				throw new Exception('Cube is not an rpm.class: name = ' + targetCube.name +  '.')
+		if (!targetCube.name.startsWith(RPM_CLASS)) {
+				throw new IllegalStateException('Cube is not an rpm.class: name = ' + targetCube.name +  '.')
 		}
```
bheekin: Prior to the merge, the visualizer groovy code was called via the cube tooling.visualize. In that cube, we were checking that the start cube name starts with rpm.class or rpm.enum. If not, we posted an error back to the user saying "Unable to display a visual. The visualizer currently only supports rpm.class and rpm.enum cubes."  However, now that we are calling the visualizer groovy code straight from the controller, we do need this check in the groovy code, and we also need it moved so it's performed first thing in the execution (before checking for missing scope).

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -225,7 +216,7 @@ class Visualizer extends NCubeGroovyExpression
...
 	}
 
-	private void processEnumCube (Map stack, Map stackInfo, Set visited, List nodes, List edges, Map groups, Map levels, List messages)
+	private void processEnumCube ()
 	{
```
bheekin: It's very nice to see those long argument lists gonzo.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
bheekin: I think we can in fact remove the entire if statement. It's in there from when we were displaying the traits in a long string next to each field in the tooltip. Since we are no longer doing that the code can be removed.

jsnyder4: Verified we still have to check for CLASS_TRAITS.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
jsnyder4: This method is required until we start setting required scope on all class definitions. Currently, we don't set it at all.

PR #174, opened 2016-03-09
Add scope builder to visualizer.

tympollack:


src/main/webapp/js/visualize.js
```diff
@@ -781,6 +798,139 @@ var Visualizer = (function ($) {
...
+        var newScope = buildScopeText();
+        _scopeInput.val(newScope);
+        _scope = buildScopeFromText(newScope);
+        load(true);
+        $(_scopeBuilderModal).modal('hide');
+    }
+
+    // TODO - extract to cube
+    var availableScopeKeys = ['context','action','state','date','product','coverage','risk','classCode','programType','lob','businessUnit','producer'];
+    function addNewScopeBuilderTableRow() {
```
bheekin: The following additional scope keys are used in the old WC model for state/date scoping. We should probably add to the list? 

policyControlDate, quoteDate, LocationState

PR #178, opened 2016-03-11
Think all search issues have been fleshed out.

tympollack:


src/main/webapp/js/ntwobe.js
```diff
@@ -459,31 +473,148 @@ var NCubeEditor2 = (function ($) {
...
+                        axisOrder.push(getAxisId(axes[axisIdx]));
+                    }
+                    for (var cIdx = 0, cLen = _columnIdCombinationsToShow.length; cIdx < cLen; cIdx++) {
+                        var combo = _columnIdCombinationsToShow[cIdx];
+                        var colId = axisColumnMap[axes[axisNum].name][colNum];
+                        var charIdx = combo.indexOf(colId);
+                        if (charIdx > -1) {
+                            var add = cIdx === 0;
+                            if (!add) {
+                                var getAxesOrderedId = function(numericallyOrderedString) {
```
jdereg: Seems like this function (getAxesOrderedId()) should be brought up to a higher level as it could get used again, and it would help reduce the length of the getColumnTableCoords() method.

PR #187, opened 2016-03-17
Introduce batch update for reference axes. (synchronous version)

tympollack:


src/main/webapp/index.html
```diff
@@ -577,10 +580,70 @@ <h3 id="changeVerLabel" class="modal-title"></h3>
     </div>
 </div>
 
+<!-- Delete Cube Modal -->
```
jdereg: Copy paste error.  I think the comment was supposed to say Batch Update Axis Ref Modal

PR #192, opened 2016-03-21
Fix issue where undefined would appear when editing a cell by typing …

tympollack:
…when no default cell value present.


src/main/webapp/js/ntwobe.js
```diff
@@ -2667,7 +2667,7 @@ var NCubeEditor2 = (function ($) {
             isCached = data.defaultCellValueCache;
         }
         // Set the cell value (String)
-        var cellValue = value !== null ? value : '';
+        var cellValue = value !== null && value !== undefined ? value : '';
```
jdereg: if (value) covers both value != null && value != undefined

PR #200, opened 2016-04-01
First pass at functional and data access control.

tympollack:


jdereg:
We need to work on the sys.permissions and sys.usergroups n-cube structures.  There needs to be a cube that maps users to groups.  It is a truth table that has users on one axis and groups on the other axis.  These two axes are technically reference axes, but we do not have to make them reference axes.  Other n-cubes that need the users (if ever) or the groups would point to the axis on this users-to-groups n-cube.

The cells of this cube are simply boolean true / false depending on whether the user is in the group our not.  This approach avoids repeating user names.

A user belongs to a group if there is a true in the cell for the user/group combination. 
-----
For the sys.permissions tables, I want to look at the way resources are specified.  For example, we could match simple named resources (no brackets) and potentially use 'pathing' to allow more macro description of resources to more specific control:
/foo/bar/baz/qux
/foo/bar
In these cases, giving access to the resource /foo/bar gives access to /foo/bar/baz/qux.  Lot's more to discuss, but I want to get these definitions ironed out first.

PR #201, opened 2016-04-04
See commit messages.

tympollack:


src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
```diff
@@ -129,16 +139,82 @@ class NCubeController extends BaseController
...
+            ApplicationID bootVersion = new ApplicationID(appId.tenant, appId.app, '0.0.0', ReleaseStatus.SNAPSHOT.toString(), ApplicationID.HEAD)
+            NCube permCube = nCubeService.getCube(bootVersion, 'sys.permissions')
+            NCube userCube = nCubeService.getCube(bootVersion, 'sys.usergroups')
+
+            // default permission for action against resource
+            boolean hasPermission = permCube.getCell(['resource': resource, 'action': action, 'group': null])
+
+            // handle exceptions
+            Axis groupAxis = permCube.getAxis('group')
+            Boolean permissionFromGroup = null
```
jdereg: boolean  permissionFromGroup = false instead of Boolean

src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
```diff
@@ -129,16 +139,82 @@ class NCubeController extends BaseController
...
+            for (Column groupColumn : groupAxis.getColumnsWithoutDefault())
+            {
+                String colName = groupColumn.getValue()
+                Boolean isGroupActive = permCube.getCell(['resource': resource, 'action': action, 'group': colName])
+                if (isGroupActive != null && isUserInGroup(userCube, colName))
+                {
+                    permissionFromGroup = permissionFromGroup || isGroupActive
```
jdereg: permissionFromGroup |= isGroupActive

src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
```diff
@@ -129,16 +139,82 @@ class NCubeController extends BaseController
...
+                }
+            }
+            if (permissionFromGroup == null)
```
jdereg: no null check needed.

src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
```diff
@@ -129,16 +139,82 @@ class NCubeController extends BaseController
...
+            {
+                return hasPermission
+            }
+            return hasPermission ^ permissionFromGroup
```
jdereg: return hasPermission & permissionFromGroup  (why use XOR?)

src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
```diff
@@ -129,16 +139,82 @@ class NCubeController extends BaseController
...
+    private boolean isUserInGroup(NCube userCube, String groupName)
+    {
+        boolean defaultInGroup = userCube.getCell(['role': groupName, 'users': null])
+        boolean isException = userCube.getCell(['role': groupName, 'users': getUser()])
+
+        if (isException == null)
+        {
+            return defaultInGroup
+        }
+        return defaultInGroup ^ isException
```
jdereg: Why use xor?  I would think 'and'

src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
tympollack: Use Boolean bc not all cells are filled.

src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
tympollack: Yes if not all cells filled.

src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
tympollack: the xors made more sense when it was a black/white list.  I can change it.

src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
```diff
@@ -577,7 +656,7 @@ class NCubeController extends BaseController
         try
         {
             appId = addTenant(appId)
-            isAllowed(appId, cubeName, Delta.Type.ADD)
+            isAllowed(appId, null, ACTION.ADD)
```
jdereg: What is a user tries to add 'sys.classpath', 'sys.prototype', 'sys.menu' etc.  Will this still be caught if the cube name is not checked here?

src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
tympollack: Scratch that.  It's still the same.  If you use hasPermission && permission from group, and say default is false but this group is true, it will still be false.  Can't do || bc if the default is true and a group would be false, would still be true.

src/main/webapp/js/constants.js
```diff
@@ -124,6 +124,11 @@ var DEFAULT_SCOPE = [
     }
 ];
 
+var PERMISSION_ACTION = {
+    RELEASE: 'release',
+    COMMIT: 'commit'
```
jdereg: I would think 'ROLLBACK' needs to be added as an action?

src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
jdereg: All cells should be filled - use n-cube level default.  It would be invalid for not all cells to be populated.

src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
jdereg: all cells should be populated in permissions T/F tables

src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
jdereg: & should be correct and the most readable.

src/main/webapp/js/constants.js
tympollack: That was not in the requirements.

src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
tympollack: No, but in that case, both null then cubename would need to be checked. So it's wrong either way.

src/main/webapp/js/constants.js
jdereg: The requirements are 'what makes sense'.  I am asking that if we are going to have permissions specific to commit and release, I think we also need rollback.

src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
tympollack: & is not correct for the same reason stated above.

jdereg: Let's talk more on this.  My (subject = me) permission on some action against a resource, is the result of ORing (checking) all of the groups I belong to.  If any are true, then I have access.

tympollack: That would be true if there weren't a default.  The default allows us to specify a white or black list.  Then if we have 30 groups and only admin needs permission, we say default false admin true instead of putting false for 29 groups.

jdereg: That's why we should be using a default cell value of boolean false.

tympollack: But that argument stands for TRUE as well.

jdereg: For permissions, the default is FALSE.  We grant 'true' on a group basis.  This makes it easy to read / understand and is consistent with how this is done in other apps using n-cube.

tympollack: How about you modify the cube how you want it and I'll adjust.

PR #202, opened 2016-04-05
ACL Changes

tympollack:


src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
```diff
@@ -147,60 +147,106 @@ class NCubeController extends BaseController
...
-                return hasPermission
-            }
-            return hasPermission ^ permissionFromGroup
+
+            return false
         }
         catch (Exception e)
         {
-            LOG.warn("Missing " + 'sys.permissions' + " cube in the 0.0.0 version for the app: " + appId.app + ', exception ' + e.getMessage())
+            LOG.warn("Missing sys.permissions or sys.usergroups cube in the 0.0.0 version for the app: " + appId.app + ', exception ' + e.getMessage())
```
jdereg: Needs a 2nd parameter at the end (the exception 'e').  The way it is coded, only e.getMessage() is shown, losing the stack history.

src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
```diff
@@ -147,60 +147,106 @@ class NCubeController extends BaseController
...
+        }
+        if (matches.size() == 0)
+        {
+            matches << resourcePermissionAxis.getDefaultColumn()
+        }
+        return matches
+    }
 
-        if (isException == null)
+    private boolean doStringsWithWildCardsMatch(String text, String pattern)
```
jdereg: This method can be re-written in two lines, using:
{
    String regex = StringUtilities.wildcardToRegexString(wildcard)
    return text.matches(regex)
}

PR #204, opened 2016-04-12
Changes for permissions.

tympollack:


src/main/webapp/js/index.js
```diff
@@ -1490,6 +1491,11 @@ var NCE = (function ($)
         }
     }
 
+    function checkPermissions(appId, resource, action) {
+        var permissionResult = call('ncubeController.checkPermissions', [appId, resource, action]);
+        return ensureModifiable() && permissionResult.data === true;
+    }
+
```
jdereg: I really like how you have this as an 'API' instead of having to write call('ncubeController.checkPermissions', ...) everywhere.

PR #228, opened 2016-06-07
Fix multiple click listener issue.

tympollack:


src/main/webapp/js/index.js
```diff
@@ -1615,18 +1615,19 @@ var NCE = (function ($)
 
     function enableDisableLockMenu(isAppAdmin) {
         var result = call(CONTROLLER + CONTROLLER_METHOD.IS_APP_LOCKED, [getAppId()]);
+        var isLocked;
         if (result.status) {
-            setLockUnlockMenuText(result.data);
-            setGetAppLockedByMenuText();
+            setLockUnlockMenuText(isLocked);
```
jdereg: isn't 'isLocked' always undefined at this point?

tympollack: Fortunately, that's only for the first time it comes into the app (so not on every click). This is what I get for trying to declare variables at the top of scope, my bad. Also, "blonde" errors are too light to see sometimes. :) Fixed in recent commit.

PR #231, opened 2016-06-10
Set up front end for update branch; add branch cache to controller.

tympollack:


src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
```diff
@@ -42,6 +42,7 @@ import com.cedarsoftware.util.io.JsonReader
 import com.cedarsoftware.util.io.JsonWriter
 import com.google.common.util.concurrent.AtomicDouble
 import groovy.transform.CompileStatic
+import javafx.application.Application
```
jdereg: Stray import - should be removed

PR #246, opened 2016-07-08
Refactor Visualizer to use Deque for stack instead of Map

bheekin:


src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -108,24 +108,19 @@ class Visualizer extends NCubeGroovyController
...
 		NCube startCube = getCube(startCubeName)
 
-		stackInfo[TARGET_CUBE] = startCube
-		stack[stackInfo[STACK_KEY] ] = stackInfo
+		stack.push([(TARGET_CUBE): startCube, (TARGET_SCOPE): scope, (TARGET_LEVEL): 1, (EDGE_ID): 1])
 
-		while (!stack.isEmpty())
-		{
+		while(!stack.empty) {
+			cubeInfo = stack.pop()
```
jdereg: It would be much cleaner to pass cubeInfo into processCube() as opposed to relying on it as a member variable.  This makes the code closer to being ready for re-entrancy (not that it has to be).  Also, although not desired, if it were ever changed to recursive, then cubeInfo would most definitely need to be a stack based variable (passed as argument).

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -218,15 +210,15 @@ class Visualizer extends NCubeGroovyController
...
-		NCube sourceCube =  stackInfo[SOURCE_CUBE] as NCube
-		Map sourceTraitMaps = stackInfo[SOURCE_TRAIT_MAPS] as Map
-		String sourceFieldName = stackInfo[SOURCE_FIELD] as String
-		String sourceFieldRpmType =  stackInfo[SOURCE_FIELD_RPM_TYPE] as String
-		Map sourceScope = stackInfo[SOURCE_SCOPE] as Map
+		NCube sourceCube =  cubeInfo[SOURCE_CUBE] as NCube
+		Map sourceTraitMaps = cubeInfo[SOURCE_TRAIT_MAPS] as Map
+		String sourceFieldName = cubeInfo[SOURCE_FIELD] as String
+		String sourceFieldRpmType =  cubeInfo[SOURCE_FIELD_RPM_TYPE] as String
+		Map sourceScope = cubeInfo[SOURCE_SCOPE] as Map
```
jdereg: There is cost to put/getting these Map parameters on each iteration.  And the code to do this is duplicated.  If you created a class that represented this cubeInfo (which I think may be more than cubeInfo and could use a more descriptive name), then you could just push this single class (not Map) on the stack, pop it off, and work with the 1st order class.  The 'dotting' of the fields will be faster than all the Map look-ups (not crazy faster but faster).  Based on the fact that this 'cubeInfo' contains information about a source n-cube and information about a target n-cube, is this really a 'relationshipInfo' or something else?

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -237,12 +229,12 @@ class Visualizer extends NCubeGroovyController
 			return
 		}
 
-		long targetLevel = stackInfo[TARGET_LEVEL] as long
+		long targetLevel = cubeInfo[TARGET_LEVEL] as long
```
jdereg: If an explicit 'relationshipInfo' (or whatever it is called) class is used, then these casts will not be needed as you fetch the data from each member.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -280,7 +271,7 @@ class Visualizer extends NCubeGroovyController
 		}
 
 
-		addToEdges(targetCube, sourceCube, targetScope, sourceScope, targetTraitMaps, sourceTraitMaps, sourceFieldName, stackInfo[STACK_KEY] as long, targetLevel)
+		addToEdges(targetCube, sourceCube, targetScope, sourceScope, targetTraitMaps, sourceTraitMaps, sourceFieldName, cubeInfo[EDGE_ID] as long, targetLevel)
```
jdereg: below, you can just add the item to the set and check the return value of the .add() call.  It is a boolean indicating whether the item was added or not.  That will make it one call to add an item to the set instead of two 1) contains check, 2) add() as opposed to just boolean changed = add().  Then check if 'changed'.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -310,33 +301,38 @@ class Visualizer extends NCubeGroovyController
 		groups[AVAILABLE_GROUPS_ALL_LEVELS] = available.toArray()
```
jdereg: All of these calls to .toArray() and 'as Set' are very expensive.  Need to look at why all this data copying around and converting is going on.

jdereg:
Performance related comment - In the 'getDotSuffix()' method, it is obtaining the pos of value.lastIndexOf('.') and not saving it.  That is an expensive call - linearly accessing the String.  Save the value in a local, and then re-use it the return statement on the next line.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
bheekin: Hi John, thanks for all good feedback on this PR. I will try to get these changes in today while the iron is still hot. However, as for this particular comment, I'm not sure what you mean. Can you stop by when you have a moment to explain?

PR #247, opened 2016-07-11
Visualizer changes based on comments made on PR 246

bheekin:
1.	Create new VisualizerRelnfo class that holds info about a source and target cube and their relationship. Use new class in Visualizer instead of a map. 
2.	Pass in VisualizerRelnfo to the  processCube method and its sub-methods, rather than relying on as member variable.
3.	Create new Visualizerlnfo class that holds info about scope, groups, levels, nodes and edges. Use new class in Visualizer instead of maps. 
4.	Remove usage of .toArray.
5.	Create variable for value.lastIndexOf in getDotSuffix method.

### DIFF
--- a/src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
+++ b/src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
@@ -30,21 +30,13 @@ class Visualizer extends NCubeGroovyController
 	public static final String AXIS_NAME = 'name'
 	public static final String AXIS_TRAIT = 'trait'
 
-	public static final String MAX_LEVEL = 'maxLevel'
-	public static final String NODE_COUNT = 'nodeCount'
-	public static final String SELECTED_LEVEL = 'selectedLevel'
-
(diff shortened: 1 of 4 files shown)


src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -24,146 +24,123 @@ class Visualizer extends NCubeGroovyController
...
-	private List edges = []
+	public static final String STATUS_MISSING_SCOPE = 'missingScope'
+	public static final String STATUS_SUCCESS = 'success'
+
 	private List messages = []
 	private Set visited = []
-	private Map cubeInfo = [:]
-	Deque stack = new ArrayDeque<>()
+
+	Deque<VisualizerRelInfo> stack = new ArrayDeque<VisualizerRelInfo>()
```
jdereg: Deque<VisualizerRelInfo> stack = new ArrayDeque<>() - Since Java 1.7, the template name should not be restated on the right.  You will see it collapsed and grayed out in Intellij.

PR #252, opened 2016-08-03
Added getSearchCount that returns the number of results found by search

jsnyder4:
DevOps requested a way to see the active n-cube count for a given ApplicationID, which will allow them to automate the verification of an n-cube release.

I decided to piggyback off of search and default cubeNamePattern, content and active to make the call easier for them, but also allow for a more robust search count if necessary.


jsnyder4:
I had a couple different names locally, so I'm cool with whatever makes the most sense. At first I did getAppCubeCount, until I realized it could be a count based on some search parameters. Then I changed to getSearchCount since it's really just telling you how many results you got back, but getCubeCount might be more generic.

PR #271, opened 2016-09-01
Global comparator between any two cubes.

tympollack:


src/main/webapp/index.html
```diff
@@ -795,6 +796,60 @@ <h4 class="panel-title">
     </div>
 </div>
 
+<!-- Batch Update Axis References Modal -->
```
jdereg: comment incorrect - copy/paste issue

PR #288, opened 2016-10-05
Compare tool / update modal changes.

tympollack:


src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
```diff
@@ -1327,22 +1327,19 @@ class NCubeController extends BaseController
 
     Object commitCube(ApplicationID appId, String cubeName)
     {
+        appId = addTenant(appId)
```
jdereg: These 5 lines should be inside the try statement

src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
```diff
@@ -1353,18 +1350,15 @@ class NCubeController extends BaseController
 
     Object commitBranch(ApplicationID appId, Object[] infoDtos)
     {
+        appId = addTenant(appId)
```
jdereg: this statement should be inside the try statement

src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
```diff
@@ -1487,7 +1480,7 @@ class NCubeController extends BaseController
     {
         appId = addTenant(appId)
         NCube ncube = nCubeService.getCube(appId, cubeName)
-        ncube.setDefaultCellValue(null)
+        ncube.defaultCellValue = null
```
jdereg: need to verify that this is actually calling setDefaultValue.  It if is not, then this needs to be returned to using the setter API.  Otherwise, clearSha1() will not be called inside the setter API.

src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
```diff
@@ -1500,34 +1493,44 @@ class NCubeController extends BaseController
                 CellInfo.parseJsonValue(cellInfo.value, null, cellInfo.dataType, cellInfo.isCached)
 
         NCube ncube = nCubeService.getCube(appId, cubeName)
-        ncube.setDefaultCellValue(cellValue)
+        ncube.defaultCellValue = cellValue
```
jdereg: Need to verify that this calls the setDefaultValue() API which clears the SHA-1.

src/main/webapp/js/common.js
```diff
@@ -405,33 +361,42 @@ function populateSelect(nce, sel, method, params, defVal, forceRefresh, isInvert
...
 
 function populateSelectFromCube(nce, sel, params, searchType) {
+    var result, results, idx, iLen, axis;
     var axisTypes = {};
+    
+    function getSelectHtml(results) {
+        var i, len, obj, val;
+        var html = '';
+        for (i = 0, len = results.length; i < len; i++) {
+            obj = null;
```
jdereg: This assignment to null can be moved outside the for loop.

src/main/webapp/js/common.js
```diff
@@ -405,33 +361,42 @@ function populateSelect(nce, sel, method, params, defVal, forceRefresh, isInvert
...
-    if (result.status === true) {
-        var results = JSON.parse(result.data).axes;
+    result = nce.call(CONTROLLER + CONTROLLER_METHOD.GET_JSON, params, {noResolveRefs:true});
+    if (result.status) {
+        results = JSON.parse(result.data).axes;
         if (searchType === POPULATE_SELECT_FROM_CUBE.METHOD) {
-            for (var idx = 0, iLen = results.length; idx < iLen; idx++) {
-                var axis = results[idx];
+            for (idx = 0, iLen = results.length; idx < iLen; idx++) {
+                axis = null;
```
jdereg: assignment to null can be moved outside the for loop.

src/main/webapp/js/ntwobe.js
```diff
@@ -654,33 +658,39 @@ var NCubeEditor2 = (function ($) {
...
 
     function clearSearchMatches() {
-        for (var axisNum = 0, axisLen = axes.length; axisNum < axisLen; axisNum++) {
-            var axis = axes[axisNum];
-            var cols = axis.columns;
-            var colKeys = Object.keys(cols);
-            for (var colNum = 0, colLen = colKeys.length; colNum < colLen; colNum++) {
+        var axisNum, axisLen, axis, cols, colKeys, cells, cellKeys, i, len, colNum, colLen;
+        for (axisNum = 0, axisLen = axes.length; axisNum < axisLen; axisNum++) {
+            axis = null;
```
jdereg: These assignments to null should be down outside the 'for' loop.

src/main/webapp/js/ntwobe.js
```diff
@@ -3674,62 +3694,70 @@ var NCubeEditor2 = (function ($) {
...
+        }
+        axis = _columnList.prop('model');
+        html = '';
+
+        for (c = 0, cLen = addedColVals.length; c < cLen; c++){
+            addedColVal = addedColVals[c];
+            if (addedColVal === '') {
+                continue; // blank row from excel copy
+            }
+            newCol = null;
```
jdereg: assignment to null should be down outside the for loop so that the memory is reclaimed at the end.

src/main/webapp/js/prop-builder.js
```diff
@@ -126,82 +130,98 @@ var PropertyBuilder = (function ($) {
...
         var columnKeys = _options.columnKeys;
         var columns = _options.columns;
-        for (var c = 0, cLen = columnKeys.length; c < cLen; c++) {
-            var key = columnKeys[c];
-            var column = columns[key];
-            var dataVal = null;
-            var inputElement = null;
+        for (c = 0, cLen = columnKeys.length; c < cLen; c++) {
+            key = columnKeys[c];
+            column = null;
```
jdereg: assignment to null should be done outside the 'for' loop so memory is reclaimed.

src/main/webapp/js/prop-builder.js
```diff
@@ -264,32 +284,41 @@ var PropertyBuilder = (function ($) {
...
-                var type = column.type;
-                var inputElement = tr.find('.' + key);
-                var dataVal = null;
-
-                if (type === COLUMN_TYPES.CHECKBOX) {
+        trs = findTableRows();
+        columnKeys = _options.columnKeys;
+        columns = _options.columns;
+        for (trIdx = 0, trLen = trs.length; trIdx < trLen; trIdx++) {
+            tr = null;
```
jdereg: assignment to null should be done outside the 'for' loop.

src/main/webapp/js/prop-builder.js
```diff
@@ -264,32 +284,41 @@ var PropertyBuilder = (function ($) {
...
+            _data.push(getDataRow(columns, columnKeys, tr));
+        }
+    }
+
+    function getDataRow(columns, columnKeys, tr) {
+        var c, cLen, key, inputElement, dataVal;
+        var dataRow = {};
+        for (c = 0, cLen = columnKeys.length; c < cLen; c++) {
+            key = columnKeys[c];
+            inputElement = null;
```
jdereg: assignment to null should be done outside the 'for' loop.

src/main/webapp/js/common.js
tympollack: These cases where the object is set to null INSIDE the loop before setting it to another object is to clear the object for garbage collection. If I skip setting it to null and reuse the variable, the previous object continues to live in memory with no pointer, whereas setting it to null first allows it to be GC'ed. Same for other comments.

PR #296, opened 2016-11-12
Change Visualizer to retrieve only a minimal set of traits

bheekin:
Changes:

1.	Changed Visualizer to retrieve only a minimal set of traits for each class.
2.	Removed class traits from title in Visualizer.
3.	Eliminated dependency on pd:busType trait in Visualizer (the Visualizer now only relies on traits recognized by Dynamis).

Tested by comparing json for the VisInfo object returned by Visualizer before and after changes. 

The server side performance for loading the Visualizer for the large WorkersCompensationProduct has improved by 33% as a result.

### DIFF
--- a/src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
+++ b/src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
@@ -1,6 +1,7 @@
 package com.cedarsoftware.util
 
 import com.cedarsoftware.ncube.NCube
+import com.cedarsoftware.util.io.JsonWriter
 import groovy.transform.CompileStatic
 import ncube.grv.method.NCubeGroovyController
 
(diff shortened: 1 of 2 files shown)


src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -295,22 +295,24 @@ class Visualizer extends NCubeGroovyController
 		}
 	}
 
-	private static String getGroup(NCube cube, String busType)
+	private static String getNodeGroup(NCube cube, String group)
```
jdereg: If you are only using the name field on the passed in ncube (cube), perhaps the argument should be a String name, instead of an n-cube.

bheekin: Yes, good catch. The method can also be changed to use a ternary operator + I see a few other methods that can be changed to use ternary operators. Will submit those changes with my next PR. 

	private static String getNodeGroup(String cubeName, String group)
	{
		return cubeName.startsWith(RPM_ENUM) ? group + _ENUM : group
	}

PR #297, opened 2016-11-15
Visualizer changes

bheekin:
Changes to the Visualizer:

1. Updated VisualizerHelper to run with code corresponding to Dynamis 5.2.0 (includes loading traits only for fields that are r:exists = true and handling of effective version, among other things). TODO: Get rid of VisualizerHelper by adding in Dynamis via Grapes.  Will need a new Dynamis API that allows us to load only select traits. 

2. Improved prompts to the user as to what minimal scope is required to start a visualization based on which class is to be visualized.

3. Updated Visualizer to handle missing scope beyond the minimally required start scope. If there is missing scope for a class, the node for the class  appears in the visualization and the node provides information about the issue when selected (previously, the visualization would blow up with an ugly exception). TODO: Find a way to get to missing scope keys when getCell() throws an exception without looking for the constant 'does not contain all of the required scope keys' in the root cause of exception.

4. Updated Visualizer to handle any other exceptions occurring during load of fields and traits for a given class. If such an exception occurs, the node for the class  appears in the visualization and the node provides information about the issue when selected (previously, the visualization would blow up with an ugly exception).

5. Updated Visualizer to handle reference classes where the source cube points directly to the target cube via a field, but there is no scoped name corresponding to the field name on the target cube. The node for the class appears in the visualization and the node now provides information about the issue when selected (previously, the class would display, but with minimal class info and no explanation about the issue).

6. Now only getting required scope and optional scope once for each rpm class (used to do it for each “class instance” which is not necessary and a lot of overhead).

7. Removed hard coded default scope options for the options picker in Visualizer.js. The server code is providing this info now.

### DIFF
--- a/src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
+++ b/src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
@@ -78,8 +78,8 @@ class Visualizer extends NCubeGroovyController
 
 		VisualizerInfo visInfo = new VisualizerInfo()
 		visInfo.startCubeName = options.startCubeName as String
-		Map scope = options.scope as Map
-		visInfo.scope = scope == null ? new HashMap(DEFAULT_SCOPE) : scope
+		Map scope = options.scope as LinkedHashMap
+		visInfo.scope = scope == null ? new LinkedHashMap(DEFAULT_SCOPE) : scope
 		visInfo.allGroups = ALL_GROUPS_MAP
 		visInfo.availableGroupsAllLevels = []
 		visInfo.groupSuffix = _ENUM
@@ -600,7 +600,7 @@ class Visualizer extends NCubeGroovyController
  */
 	private static Map getScopeRelativeToSource(NCube targetCube, String sourceFieldRpmType, String targetFieldName, Map<String, Object> scope)
 	{
-		Map<String, Object> newScope = new HashMap(scope)
+		Map<String, Object> newScope = new LinkedHashMap(scope)
 
 		if (targetCube.name.startsWith(RPM_ENUM))
 		{
@@ -641,7 +641,7 @@ class Visualizer extends NCubeGroovyController
 		boolean minimumScopeMissing = false
 		if (!scope || scope == DEFAULT_SCOPE){
 			if (!scope){
-				visInfo.scope = new HashMap(DEFAULT_SCOPE)
+				visInfo.scope = new LinkedHashMap(DEFAULT_SCOPE)
 			}
 			String type = Splitter.on('.').split(startCubeName.replace(RPM_CLASS_DOT, '')).first()
 			visInfo.scope[type.toLowerCase()] = DEFAULT_SCOPE_VALUE
@@ -682,7 +682,7 @@ class Visualizer extends NCubeGroovyController
 		//Check other required scope for the start cube
 		Set<String> missingScope = findMissingScope(startCubeName, scope)
 		if (missingScope) {
-			Map expandedScope = scope == null ? new HashMap(DEFAULT_SCOPE) : new HashMap(scope)
+			Map expandedScope = scope == null ? new LinkedHashMap<>(DEFAULT_SCOPE) : new LinkedHashMap(scope)
 			String missingScopeString = ''
 			missingScope.each {String key ->
 				expandedScope[key] = DEFAULT_SCOPE_VALUE
--- a/src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
+++ b/src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
@@ -41,7 +41,7 @@ class Visualizer extends NCubeGroovyController
 
 	public static final String _ENUM = '_ENUM'
 	public static final String UNSPECIFIED = 'UNSPECIFIED'
-	public static final Map ALL_GROUPS_MAP = [PRODUCT:'Product', FORM:'Form', RISK:'Risk', COVERAGE:'Coverage', CONTAINER:'Container', DEDUCTIBLE:'Deductible', LIMIT:'Limit', RATE:'Rate', RATEFACTOR:'Rate Factor', PREMIUM:'Premium', PARTY:'Party', PLACE:'Place', ROLE:'Role', ROLEPLAYER:'Role Player', UNSPECIFIED:'Unspecified']
+	public static final Map ALL_GROUPS_MAP = [PRODUCT: 'Product', FORM: 'Form', RISK: 'Risk', COVERAGE: 'Coverage', CONTAINER: 'Container', DEDUCTIBLE: 'Deductible', LIMIT: 'Limit', RATE: 'Rate', RATEFACTOR: 'Rate Factor', PREMIUM: 'Premium', PARTY: 'Party', PLACE: 'Place', ROLE: 'Role', ROLEPLAYER: 'Role Player', UNSPECIFIED: 'Unspecified']
 	public static final String[] GROUPS_TO_SHOW_IN_TITLE = ['COVERAGE', 'DEDUCTIBLE', 'LIMIT', 'PREMIUM', 'PRODUCT', 'RATE', 'RATEFACTOR', 'RISK', 'ROLEPLAYER', 'ROLE']
 	public static final List DERIVED_SCOPE_KEYS = ['product', 'form', 'risk', 'coverage', 'container', 'deductible', 'limit', 'rate', 'ratefactor', 'premium', 'party', 'place', 'role', 'roleplayer']
 	public static final List DERIVED_SOURCE_SCOPE_KEYS = ['sourceProduct', 'sourceForm', 'sourceRisk', 'sourceCoverage', 'sourceContainer', 'sourceDeductible', 'sourceLimit', 'sourceRate', 'sourceRate Factor', 'sourcePremium', 'sourceParty', 'sourcePlace', 'sourceRole', 'sourceRolepayer']
@@ -63,14 +63,19 @@ class Visualizer extends NCubeGroovyController
 
 	private VisualizerHelper helper = new VisualizerHelper()
 
-	public static final String STATUS_MISSING_SCOPE = 'missingScope'
+	public static final List MANDATORY_RPM_SCOPE_KEYS = [AXIS_FIELD, AXIS_NAME, AXIS_TRAIT]
+	public static final String MISSING_SCOPE = 'missing scope'
+	public static final String UNABLE_TO_LOAD = 'unable to load'
+
+	public static final String STATUS_MISSING_START_SCOPE = 'missingStartScope'
+	public static final String NOT_ALL_REQUIRED_SCOPE_KEYS = 'does not contain all of the required scope keys:'
 	public static final String STATUS_SUCCESS = 'success'
 
 	private Set messages = []
 	private Set visited = []
 	Set missingRequiredScopeKeys = []
-	Map requiredScopeKeys = [:]
-	Map optionalScopeKeys = [:]
+	Map<String, Set> requiredScopeKeys = [:]
+	Map<String, Set> optionalScopeKeys = [:]
 	Deque<VisualizerRelInfo> stack = new ArrayDeque<>()
 
 	/**
@@ -98,9 +103,9 @@ class Visualizer extends NCubeGroovyController
 		visInfo.availableGroupsAllLevels = []
 		visInfo.groupSuffix = _ENUM
 		Set selectedGroups = options.selectedGroups as Set
-		visInfo.selectedGroups = selectedGroups  == null ? ALL_GROUPS_MAP.keySet() : selectedGroups
+		visInfo.selectedGroups = selectedGroups == null ? ALL_GROUPS_MAP.keySet() : selectedGroups
 		String selectedLevel = options.selectedLevel as String
-		visInfo.selectedLevel = selectedLevel == null ? DEFAULT_LEVEL :  Converter.convert(selectedLevel, long.class) as long
+		visInfo.selectedLevel = selectedLevel == null ? DEFAULT_LEVEL : Converter.convert(selectedLevel, long.class) as long
 		Set availableScopeKeys = options.availableScopeKeys as Set
 		visInfo.availableScopeKeys = availableScopeKeys ?: DEFAULT_AVAILABLE_SCOPE_KEYS as Set
 		visInfo.maxLevel = 1
@@ -108,16 +113,15 @@ class Visualizer extends NCubeGroovyController
 		visInfo.nodes = []
 		visInfo.edges = []
 
-		if (hasMissingMinimumScope(visInfo)){
-			messages <<  'Scope is required. Please add scope value(s) for the following scope key(s): ' + String.join(COMMA_SPACE, missingRequiredScopeKeys) + '.'
-			return [status: STATUS_MISSING_SCOPE, visInfo: visInfo, message: String.join(DOUBLE_BREAK, messages)]
+		if (hasMissingMinimumScope(visInfo)) {
+			messages << 'Scope is required. Please add scope value(s) for the following scope key(s): ' + String.join(COMMA_SPACE, missingRequiredScopeKeys) + '.'
+			return [status: STATUS_MISSING_START_SCOPE, visInfo: visInfo, message: String.join(DOUBLE_BREAK, messages)]
 		}
 
 		getRpmVisualization(visInfo)
 
 		if (missingRequiredScopeKeys) {
 			messages << 'Additional scope is required. Please add scope value(s) for the following scope key(s): ' + String.join(COMMA_SPACE, missingRequiredScopeKeys) + '.'
-			return [status: STATUS_MISSING_SCOPE, visInfo: visInfo, message: String.join(DOUBLE_BREAK, messages)]
 		}
 
 		String message = messages.size() > 0 ? String.join(DOUBLE_BREAK, messages) : null
@@ -127,13 +131,13 @@ class Visualizer extends NCubeGroovyController
 	private void getRpmVisualization(VisualizerInfo visInfo)
 	{
 		VisualizerRelInfo relInfo = new VisualizerRelInfo()
-		relInfo.targetCube =  getCube(visInfo.startCubeName)
+		relInfo.targetCube = getCube(visInfo.startCubeName)
 		relInfo.targetScope = visInfo.scope
 		relInfo.targetLevel = 1
 		relInfo.id = 1
 		stack.push(relInfo)
 
-		while(!stack.empty) {
+		while (!stack.empty) {
 			processCube(visInfo, stack.pop())
 		}
 
@@ -144,8 +148,8 @@ class Visualizer extends NCubeGroovyController
 		trimSelectedGroups(visInfo)
 	}
 
-	private static Set addSets(Set<String> set, Set<Set> sets){
-		sets.each{
+	private static Set addSets(Set<String> set, Set<Set> sets) {
+		sets.each {
 			set.addAll(it)
 		}
 	}
@@ -166,12 +170,12 @@ class Visualizer extends NCubeGroovyController
 	{
 		String targetCubeName = relInfo.targetCube.name
 		Map targetScope = relInfo.targetScope
+		Map targetTraitMaps = [:]
 
-		boolean canLoadTargetAsRpmClass = canLoadTargetAsRpmClass(relInfo)
-		boolean targetHasMissingRequiredScopeKeys = hasMissingRequiredScopeKeys(visInfo, relInfo)
-		if (canLoadTargetAsRpmClass && !targetHasMissingRequiredScopeKeys)
+		boolean loadFieldsAndTraits = canLoadTargetAsRpmClass(relInfo)
+		if (loadFieldsAndTraits)
 		{
-			relInfo.targetTraitMaps = helper.getTraitMaps(targetCubeName, targetScope)
+			loadFieldsAndTraits = getTraitMaps(visInfo, relInfo, targetTraitMaps)
 		}
 
 		addToEdges(visInfo, relInfo)
@@ -185,40 +189,38 @@ class Visualizer extends NCubeGroovyController
 		visInfo.availableGroupsAllLevels << group
 		addToNodes(visInfo, relInfo, group)
 
-		if (!canLoadTargetAsRpmClass || targetHasMissingRequiredScopeKeys){
-			return
-		}
-
-		Map targetTraitMaps = relInfo.targetTraitMaps
-		targetTraitMaps.each{ k, v ->
-			String targetFieldName = k as String
-			Map targetTraits = v as Map
-			Boolean exists = targetTraits[R_EXISTS]
-			if (!CLASS_TRAITS.equals(targetFieldName) && exists)
-			{
-				String targetFieldRpmType = targetTraits[R_RPM_TYPE]
-
-				if (!helper.isPrimitive(targetFieldRpmType))
+		if (loadFieldsAndTraits)
+		{
+			targetTraitMaps.each { k, v ->
+				String targetFieldName = k as String
+				Map targetTraits = v as Map
+				Boolean exists = targetTraits[R_EXISTS]
+				if (!CLASS_TRAITS.equals(targetFieldName) && exists)
 				{
-					NCube nextTargetCube
-					String nextTargetCubeName
-					if (targetTraits.containsKey(V_ENUM))
-					{
-						nextTargetCubeName = RPM_ENUM_DOT + targetTraits[V_ENUM]
-						nextTargetCube = getCube(nextTargetCubeName)
-					}
-					else if (targetFieldRpmType)
-					{
-						nextTargetCubeName = RPM_CLASS_DOT + targetFieldRpmType
-						nextTargetCube = getCube(nextTargetCubeName)
-					}
+					String targetFieldRpmType = targetTraits[R_RPM_TYPE]
 
-					if (nextTargetCube)
+					if (!helper.isPrimitive(targetFieldRpmType))
 					{
-						addToStack(visInfo, relInfo, nextTargetCube, targetFieldRpmType, targetFieldName)
-					}
-					else{
-						messages << 'No cube exists with name of ' + nextTargetCubeName + '. It is therefore not included in the visualization.'
+						NCube nextTargetCube
+						String nextTargetCubeName
+						if (targetTraits.containsKey(V_ENUM))
+						{
+							nextTargetCubeName = RPM_ENUM_DOT + targetTraits[V_ENUM]
+							nextTargetCube = getCube(nextTargetCubeName)
+						}
+						else if (targetFieldRpmType)
+						{
+							nextTargetCubeName = RPM_CLASS_DOT + targetFieldRpmType
+							nextTargetCube = getCube(nextTargetCubeName)
+						}
+
+						if (nextTargetCube)
+						{
+							addToStack(visInfo, relInfo, nextTargetCube, targetFieldRpmType, targetFieldName)
+						}
+						else {
+							messages << 'No cube exists with name of ' + nextTargetCubeName + '. It is therefore not included in the visualization.'
+						}
 					}
 				}
 			}
@@ -227,25 +229,24 @@ class Visualizer extends NCubeGroovyController
 
 	private void processEnumCube(VisualizerInfo visInfo, VisualizerRelInfo relInfo)
 	{
+		String group = UNSPECIFIED
+		Map targetTraitMaps = [:]
 		Map targetScope = relInfo.targetScope
 		String targetCubeName = relInfo.targetCube.name
-		String sourceFieldRpmType =  relInfo.sourceFieldRpmType
+		String sourceFieldRpmType = relInfo.sourceFieldRpmType
 
 		if (!targetCubeName.startsWith(RPM_ENUM)) {
-			throw new IllegalStateException('Cube is not an rpm.enum cube: ' + targetCubeName +  '.')
+			throw new IllegalStateException('Cube is not an rpm.enum cube: ' + targetCubeName + '.')
 		}
 
-		if(relInfo.sourceCube && (!sourceFieldRpmType || helper.isPrimitive(sourceFieldRpmType)))
+		if (relInfo.sourceCube && (!sourceFieldRpmType || helper.isPrimitive(sourceFieldRpmType)))
 		{
 			return
 		}
 
-		String group = UNSPECIFIED
+		boolean loadFieldsAndTraits = getTraitMaps(visInfo, relInfo, targetTraitMaps)
 
-		boolean targetHasMissingRequiredScopeKeys = hasMissingRequiredScopeKeys(visInfo, relInfo)
-		if (!targetHasMissingRequiredScopeKeys) {
-			relInfo.targetTraitMaps = helper.getTraitMaps(targetCubeName, targetScope)
-			Map targetTraitMaps = relInfo.targetTraitMaps
+		if (loadFieldsAndTraits) {
 			targetTraitMaps.each { k, v ->
 				String targetFieldName = k as String
 				Map targetTraits = v as Map
@@ -350,11 +351,11 @@ class Visualizer extends NCubeGroovyController
 			List columns = classTraitsCube.getAxis(type.toLowerCase()).columns
 			String sourceFieldName = relInfo.sourceFieldName
 			if (!columns.find { sourceFieldName == (it as Column).value }) {
-				relInfo.targetTraitMaps = [(CLASS_TRAITS):[(R_SCOPED_NAME):'n/a']]
+				relInfo.targetTraitMaps = [(CLASS_TRAITS): [(R_SCOPED_NAME): UNABLE_TO_LOAD]]
 				relInfo.notes << 'The source cube ' + relInfo.sourceCube.name + ' points directly to this cube (' + targetCube.name +
 						') via field ' + sourceFieldName + ', but there is no ' + type.toLowerCase() + ' named ' +
 						sourceFieldName + ' on this cube. It can therefore not be loaded as an rpm.class in the visualization.'
-				relInfo.loadTargetAsRpmClass = false
+				relInfo.loadFieldsAndTraits = false
 				return false
 			}
 		}
@@ -374,7 +375,7 @@ class Visualizer extends NCubeGroovyController
 		return groups.contains(group) ? group : UNSPECIFIED
 	}
 
-	private static void addToEdges(VisualizerInfo visInfo, VisualizerRelInfo relInfo )
+	private static void addToEdges(VisualizerInfo visInfo, VisualizerRelInfo relInfo)
 	{
 		NCube sourceCube = relInfo.sourceCube
 		if (!sourceCube)
@@ -403,7 +404,7 @@ class Visualizer extends NCubeGroovyController
 		if (vMin != null && vMax != null) {
 			edgeMap.title = vMin + ':' + vMax
 		}
-		else{
+		else {
 			edgeMap.title = ''
 		}
 		visInfo.edges << edgeMap
@@ -462,9 +463,9 @@ class Visualizer extends NCubeGroovyController
 		visInfo.nodes << nodeMap
 	}
 
-	private static String getTitle(VisualizerRelInfo relInfo, Map nodeMap)
+	private String getTitle(VisualizerRelInfo relInfo, Map nodeMap)
 	{
-		boolean loadAsRpmClass = relInfo.loadTargetAsRpmClass
+		boolean loadFieldsAndTraits = relInfo.loadFieldsAndTraits
 		Map scope = relInfo.targetScope
 		Map traitMaps = relInfo.targetTraitMaps
 		Set notes = relInfo.notes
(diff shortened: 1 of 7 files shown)


src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -72,50 +97,63 @@ class Visualizer extends NCubeGroovyController
 
 		VisualizerInfo visInfo = new VisualizerInfo()
 		visInfo.startCubeName = options.startCubeName as String
-		Map scope = options.scope as Map
-		visInfo.scope = scope == null ? DEFAULT_SCOPE : scope
+		Map scope = options.scope as LinkedHashMap
+		visInfo.scope = scope == null ? new LinkedHashMap(DEFAULT_SCOPE) : scope
```
jdereg: visInfo.scope = scope ?: new LinkedHashMap(DEFAULT_SCOPE)

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -72,50 +97,63 @@ class Visualizer extends NCubeGroovyController
...
 		visInfo.allGroups = ALL_GROUPS_MAP
 		visInfo.availableGroupsAllLevels = []
 		visInfo.groupSuffix = _ENUM
 		Set selectedGroups = options.selectedGroups as Set
-		visInfo.selectedGroups = selectedGroups  == null ? ALL_GROUPS_MAP.keySet() : selectedGroups
+		visInfo.selectedGroups = selectedGroups == null ? ALL_GROUPS_MAP.keySet() : selectedGroups
```
jdereg: you can use Elvis operator here

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -72,50 +97,63 @@ class Visualizer extends NCubeGroovyController
...
 		visInfo.nodes = []
 		visInfo.edges = []
 
-		//TODO: This is a temp location for checking scope. Move check for scope to where traitMaps are obtained.
-		//TODO: Also handle the user adding the required scope key but with a value that results in a no hit.
-		if (hasMissingScope(visInfo))
-		{
-			return [status: STATUS_MISSING_SCOPE, visInfo: visInfo, message: messages.toString()]
+		if (hasMissingMinimumScope(visInfo)) {
+			messages << 'Scope is required. Please add scope value(s) for the following scope key(s): ' + BREAK + BREAK + String.join(COMMA_SPACE, missingRequiredScopeKeys)
```
jdereg: Using double-quoted (") Strings will allow you to put the variables in-line in the String and remove the String concatenation.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -72,50 +97,63 @@ class Visualizer extends NCubeGroovyController
...
+			return [status: STATUS_MISSING_START_SCOPE, visInfo: visInfo, message: String.join(DOUBLE_BREAK, messages)]
 		}
 
 		getRpmVisualization(visInfo)
 
-		String message = messages.size() > 0 ? messages.toString() : null
+		if (missingRequiredScopeKeys) {
+			messages << 'Additional scope is required. Please add scope value(s) for the following scope key(s): ' + BREAK + BREAK + String.join(COMMA_SPACE, missingRequiredScopeKeys)
```
jdereg: Use double (") strings instead of String concatenation (+).  "This ${var} is easier to${read}" than 'This ' + var.toString() + ' is easier to ' + read.toString().

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -132,51 +170,57 @@ class Visualizer extends NCubeGroovyController
...
+							nextTargetCubeName = RPM_CLASS_DOT + targetFieldRpmType
+							nextTargetCube = getCube(nextTargetCubeName)
+						}
+
+						if (nextTargetCube)
+						{
+							addToStack(visInfo, relInfo, nextTargetCube, targetFieldRpmType, targetFieldName)
+						}
+						else {
+							messages << 'No cube exists with name of ' + nextTargetCubeName + '. It is therefore not included in the visualization.'
```
jdereg: use double-quoted strings

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -185,56 +229,54 @@ class Visualizer extends NCubeGroovyController
...
+		String group = UNSPECIFIED
+		Map targetTraitMaps = [:]
 		Map targetScope = relInfo.targetScope
 		String targetCubeName = relInfo.targetCube.name
-		String sourceFieldRpmType =  relInfo.sourceFieldRpmType
+		String sourceFieldRpmType = relInfo.sourceFieldRpmType
 
 		if (!targetCubeName.startsWith(RPM_ENUM)) {
-			throw new IllegalStateException('Cube is not an rpm.enum cube: ' + targetCubeName +  '.')
+			throw new IllegalStateException('Cube is not an rpm.enum cube: ' + targetCubeName + '.')
```
jdereg: use double quoted strings

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -185,56 +229,54 @@ class Visualizer extends NCubeGroovyController
...
+							if (nextTargetCube)
+							{
+								addToStack(visInfo, relInfo, nextTargetCube, relInfo.sourceFieldRpmType, targetFieldName)
+
+								if (group == UNSPECIFIED) {
+									group = getGroup(nextTargetCubeName)
+								}
+							}
+							else {
+								messages << 'No cube exists with name of ' + nextTargetCubeName + '. It is therefore not included in the visualization.'
```
jdereg: use double quoted strings

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -185,56 +229,54 @@ class Visualizer extends NCubeGroovyController
...
-							messages << 'No cube exists with name of ' + nextTargetCubeName + '. It is therefore not included in the visualization.'
 						}
 					}
-				}
-				catch (Exception e)
-				{
-					throw new IllegalStateException('Exception caught while loading and processing the cube for enum field ' + targetFieldName + ' in enum ' + targetCubeName + '.', e)
+					catch (Exception e)
+					{
+						throw new IllegalStateException('Exception caught while loading and processing the cube for enum field ' + targetFieldName + ' in enum ' + targetCubeName + '.', e)
```
jdereg: use double quoted strings

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -295,16 +335,36 @@ class Visualizer extends NCubeGroovyController
...
+		NCube sourceCube = relInfo.sourceCube
+		NCube targetCube = relInfo.targetCube
+
+		if (sourceCube && sourceCube.name.startsWith(RPM_CLASS_DOT) && targetCube.name.startsWith(RPM_CLASS_DOT) &&
+				targetCube.getAxis(AXIS_TRAIT).contains(R_SCOPED_NAME))
 		{
-			return group
+			String type = Splitter.on('.').split(targetCube.name.replace(RPM_CLASS_DOT, '')).first()
+			NCube classTraitsCube = getCube(RPM_SCOPE_CLASS_DOT + type + DOT_CLASS_TRAITS)
+			List columns = classTraitsCube.getAxis(type.toLowerCase()).columns
```
jdereg: No need to call type.toLowerCase() when getting axis name.  That is an expensive call, and the Axes are compared case-insensitively already.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -295,16 +335,36 @@ class Visualizer extends NCubeGroovyController
...
+			String sourceFieldName = relInfo.sourceFieldName
+			if (!columns.find { sourceFieldName == (it as Column).value }) {
+				relInfo.targetTraitMaps = [(CLASS_TRAITS): [(R_SCOPED_NAME): UNABLE_TO_LOAD]]
+				relInfo.notes << 'The source cube ' + relInfo.sourceCube.name + ' points directly to this cube (' + targetCube.name +
```
jdereg: use double " strings (or triple double quotes if you need to use " and ' in the String)

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -398,55 +458,72 @@ class Visualizer extends NCubeGroovyController
...
+		Set notes = relInfo.notes
+		String scopedName = getScopedName(traitMaps)
+
 		StringBuilder sb = new StringBuilder()
 
-		if (scopedName != null )
-		{
-			sb.append('<strong>scoped name = </strong>' + scopedName + '<br><br>')
+		if (scopedName) {
+			sb.append('<strong>scoped name = </strong>' + scopedName + DOUBLE_BREAK)
```
jdereg: "b" is shorter than "strong" and is now accepted for use.  See absolute latest HTML specs.
Also, you can use double " string here instead of concatenation (+)

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -398,55 +458,72 @@ class Visualizer extends NCubeGroovyController
...
 		}
 
-		sb.append('<strong>scope = </strong>' + scope.toString().replace('{','').replace('}','') + '<br><br>')
+		if (notes)
+		{
+			sb.append('<strong>Note: </strong><br>')
```
jdereg: "b" is shorter than "strong"
use double quote " string

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -398,55 +458,72 @@ class Visualizer extends NCubeGroovyController
...
+			notes.each { String note ->
+				sb.append(' ' + note + BREAK)
+			}
+			sb.append(BREAK)
+		}
 
-		String requiredScope = getRequiredScopeAllReferenced(targetCube).toString().replace('[','').replace(']','')
-		sb.append(requiredScope == '' ? '' : '' + '<strong>required scope = </strong>' + requiredScope  + '<br><br>')
+		sb.append('<strong>scope = </strong>' + scope.toString().replace('{', '').replace('}', '') + DOUBLE_BREAK)
```
jdereg: Since there is already a StringBuilder being used, no need to do String concatenation inside of it.  Use the .append() method, or use the double " string.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -398,55 +458,72 @@ class Visualizer extends NCubeGroovyController
...
 
-		String optionalScope = getOptionalScopeAllReferenced(targetCube).toString().replace('[','').replace(']','')
-		sb.append(optionalScope == '' ? '' : '' + '<strong>optional scope = </strong>' + optionalScope  + '<br><br>')
+		getRequiredAndOptionalScopeKeys(relInfo)
+		String cubeName = relInfo.targetCube.name
+		String requiredScope = String.join(COMMA_SPACE, requiredScopeKeys[cubeName])
+		sb.append(requiredScope == '' ? '' : '' + '<strong>required scope to access all cells for cube = </strong>' + requiredScope + DOUBLE_BREAK)
```
jdereg: String clean up

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -398,55 +458,72 @@ class Visualizer extends NCubeGroovyController
...
-		return sb.toString();
+		return sb.toString()
 	}
 
-
 	private static String getTitleFields(Map traitMaps)
 	{
-		Map fields = getFields(traitMaps)
-		String fieldString = "<strong>fields = </strong>";
+		List<String> fields = new ArrayList(traitMaps.keySet())
```
jdereg: Using a List for the Set strings and then removing one from the list (linear operation) can be improved.
Use Set<String> fields = LinkedHashSet(traitMaps.keySet())
fields.remove(CLASS_TRAITS) // This will operate in O(1) not O(n)

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -398,55 +458,72 @@ class Visualizer extends NCubeGroovyController
...
+		fields.remove(CLASS_TRAITS)
+		String fieldString = "<strong>fields = </strong>"
```
jdereg: "b" is smaller than "strong" (faster and accepted)

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -398,55 +458,72 @@ class Visualizer extends NCubeGroovyController
...
+	private static String getFieldDetails(List<String> fields, String spaces)
 	{
 		StringBuilder fieldDetails = new StringBuilder()
 		fieldDetails.append('')
 
-		fields.each() { key, value ->
-
-			fieldDetails.append('<br>' + spaces + key + SPACE)
+		fields.each() {
+			fieldDetails.append(BREAK + spaces + it + SPACE)
```
jdereg: Doing String concatenation in a StringBuilder is inefficient.  Multiple .append() calls is faster, or use a Groovy double " string.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -560,69 +597,163 @@ class Visualizer extends NCubeGroovyController
...
-				missingScopeString = missingScopeString + key + ', '
+
+		if (!scope || scope == DEFAULT_SCOPE) {
+			if (!scope) {
+				visInfo.scope = new LinkedHashMap(DEFAULT_SCOPE)
 			}
-			missingScopeString = missingScopeString.substring(0, missingScopeString.size() - 2)
-			messages << 'Loading the visualization for ' + startCubeName + ' requires additional scope to be provided. Please add scope value(s) for the following scope key(s): ' + missingScopeString + '.'
-			visInfo.scope = expandedScope
+			String type = Splitter.on('.').split(startCubeName.replace(RPM_CLASS_DOT, '')).first().toLowerCase()
```
jdereg: This is a whole mess of String work going on here.  I would put this in a separate method so your outer logic reads cleaner.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -560,69 +597,163 @@ class Visualizer extends NCubeGroovyController
...
-			if (scope == null || !scope.containsKey(it)) {
-				missingScope << it
+	private void getRequiredAndOptionalScopeKeys(VisualizerRelInfo relInfo)
+	{
+		String cubeName = relInfo.targetCube.name
+		if (!requiredScopeKeys.containsKey(cubeName))
+		{
+			if (isRpmClassOrScopeCube(cubeName))
+			{
+				List<Set> scopeKeys = getScopeAllReferenced(relInfo.targetCube, [[] as Set, [] as Set])
```
jdereg: List.first() and List.last() can be slow operation.  Could a different List type be used here for added speed (LinkedList?)

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -560,69 +597,163 @@ class Visualizer extends NCubeGroovyController
...
+				requiredScopeKeys[cubeName] = [] as Set
+				optionalScopeKeys[cubeName] = [] as Set
 			}
 		}
+	}
 
-		return missingScope.size() > 0 ? missingScope : null
+	private static boolean isRpmClassOrScopeCube(String cubeName)
+	{
+		String cubeNamePrefix = Splitter.on(DOT_CLASS_DOT).split(cubeName).first()
```
jdereg: If we are repeat the extract of RpmClass names from n-cube names in more than one place, I would consolidate this work to one place so that the main logic is not cluttered with this work.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -560,69 +597,163 @@ class Visualizer extends NCubeGroovyController
...
+				if (missingScope)
+				{
+					Map expandedScope = scope == null ? new LinkedHashMap(DEFAULT_SCOPE) : new LinkedHashMap(scope)
+					missingScope.each { String key ->
+						expandedScope[key] = DEFAULT_SCOPE_VALUE
+					}
+					missingRequiredScopeKeys.addAll(missingScope)
+					visInfo.scope = expandedScope
+					scopedName = MISSING_SCOPE
+					notes << 'Additional scope is required. Please add scope value(s) for the following scope key(s):' + BREAK + BREAK + String.join(COMMA_SPACE, missingScope)
```
jdereg: use double " strings

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -560,69 +597,163 @@ class Visualizer extends NCubeGroovyController
...
+				}
+				else
+				{
+					throw new IllegalStateException('Exception message contains "' + NOT_ALL_REQUIRED_SCOPE_KEYS + '", but no missing keys found for ' + cubeName + ' and scope ' + scope.toString() + '.', e)
```
jdereg: use double " string

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -560,69 +597,163 @@ class Visualizer extends NCubeGroovyController
...
+				}
+			}
+			else
+			{
+				scopedName = UNABLE_TO_LOAD
+				String message = 'An exception was thrown while loading fields and traits for cube ' + cubeName + '.' + BREAK + BREAK + '<strong>Message:</strong> ' + BREAK + BREAK + e.message + BREAK + BREAK +  '<strong>Root cause: </strong>'+ BREAK + BREAK + rootCause
```
jdereg: Use double " string.
Use "b" instead of "strong"

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -560,69 +597,163 @@ class Visualizer extends NCubeGroovyController
...
+			traitMaps = [(CLASS_TRAITS): [(R_SCOPED_NAME): scopedName]]
+			relInfo.targetTraitMaps = traitMaps
+			relInfo.loadFieldsAndTraits = false
+			relInfo.notes = notes
+			return false
+		}
+		return true
+	}
+
+	private static Throwable getRootCause(Throwable cause) {
```
jdereg: This can be written more succinctly - and already exists in code elsewhere:

    static Throwable getDeepestException(Throwable e)
    {
        while (e.cause != null)
        {
            e = e.cause
        }
        return e
    }

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -560,69 +597,163 @@ class Visualizer extends NCubeGroovyController
...
+		return rootCause;
 	}
 
-	private Set getOptionalScopeAllReferenced(NCube startCube) {
-		Set optionalScope = startCube.getOptionalScope([:] as Map, [:] as Map)
-		startCube.referencedCubeNames.each {
+	private static Set<String> findMissingScope(Map scope, String requiredScopeKeyString)
+	{
+		Set<String> missingScope = []
+		String part1 = Splitter.on(NOT_ALL_REQUIRED_SCOPE_KEYS).split(requiredScopeKeyString).last()
```
jdereg: If you make methods out of all this String processing, the logic of what is actually being done will be much clearer.

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -31,35 +31,31 @@ public class VisualizerHelper extends NCubeGroovyController {
...
-	 *  COPIED: Copied from Dynamis
-	 */
 	public static
 	final Pattern PATTERN_CLASS_EXTENDS_TRAIT = Pattern.compile('[^,\\s][^\\,]*[^,\\s]*', Pattern.CASE_INSENSITIVE);
 
-	/** pattern to match r:extends trait on fields
-	 *  COPIED: Copied from Dynamis
-	 */
 	public static
 	final Pattern PATTERN_FIELD_EXTENDS_TRAIT = Pattern.compile('^\\s*((?:[a-z][a-z0-9_]*)(?:\\.[a-z][a-z0-9_]*)*)\\s*(?:[\\[]\\s*([a-z0-9_]+?)\\s*[\\]])?\\s*$', Pattern.CASE_INSENSITIVE);
```
jdereg: These should be done as groovy Patterns.  See Regex.groovy in the NCUBE code base to see how to declare regular expressions.  It removes all the extra back slashing so that the Regex is left correctly in the code.

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -114,77 +102,93 @@ public class VisualizerHelper extends NCubeGroovyController {
...
-				traits = new LinkedHashMap<String, Object>();
-				traitMaps.put(fieldName, traits);
+		// populate initial fields
+		populateAllFieldsFromAxis(fieldAndTraits, axisName, classCube);
+
+		// wildcard the fieldAxis using r:exists
+		populateExistsTrait(className, axisName, scope, fieldAndTraits, classCube, output);
+
+		// determine traits to fetch, except for r:exists already fetched
+		List<String> traitNames = getTraitNamesForCube(classCube, (String) scope.get(EFFECTIVE_VERSION_SCOPE_KEY));
```
jdereg: Semi-colons all over the place.  The code reads clearer without them.

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -114,77 +102,93 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+				continue;
+			}
+			else if (fieldAxis.findColumn(fieldName)==null) {
+				continue;
 			}
 
-			Map<String, Object> traitCells = mergeFieldTraits(fieldName, ncube,	axisName, scope, traits);
+			// gather traits for current field that haven't already been populated
+			Map<String, Object> coord = new HashMap<>(scope);
+			coord.put(axisName,fieldName);
```
jdereg: opportunity to use groovy short hand for assigning value to key in map.

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -114,77 +102,93 @@ public class VisualizerHelper extends NCubeGroovyController {
...
-			{
-				//MODIFIED: Use local CLASS_TRAITS constant
-				if (CLASS_TRAITS.equals(fieldName))
-				{
-					processClassMixins(className, extendsValue.toString(), classesToProcess);
-				}
-				else
+			// eliminate scoped fields
+			if (!isFieldValidSince(fieldTraits,(String) scope.get(EFFECTIVE_VERSION_SCOPE_KEY))) {
+				fieldTraits.put(R_EXISTS,false);
```
jdereg: opportunity for groovy-map short hand

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -114,77 +102,93 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+			}
+			if (!isFieldValidObsolete(fieldTraits,(String) scope.get(EFFECTIVE_VERSION_SCOPE_KEY))) {
+				fieldTraits.put(R_EXISTS,false);
```
jdereg: opportunity for groovy-map short hand

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -114,77 +102,93 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+			}
+
+			// check extends value
+			if (traitNames.contains(R_EXTENDS)) {
+				coord.put(TRAIT_AXIS, R_EXTENDS);
```
jdereg: opportunity for groovy-map short hand

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -114,77 +102,93 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+				Object extendsValue = classCube.getCell(coord, new HashMap(), NOT_DEFINED);
```
jdereg: instead of new HashMap(), you can just inline [:]

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -114,77 +102,93 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+				if (extendsValue!=null && hasValue(extendsValue))
 				{
-					processMasterDefinition(className, fieldName, extendsValue.toString(), cubeType, axisName, scope, traits);
+					if (!fieldTraits.containsKey(R_EXTENDS)) {
+						fieldTraits.put(R_EXTENDS,extendsValue.toString());
+					}
+
+					if (CLASS_TRAITS.equals(fieldName)) {
```
jdereg: CLASS_TRAITS == fieldName

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -229,72 +232,45 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+			boolean validField = masterCube.getAxis(axisName).findColumn(masterField) != null;
 			if (!validField) {
 				throw new IllegalArgumentException(String.format(exceptionFormat, "Field in master definition not found", fieldName,
 						classType, className, Arrays.toString(visited.toArray())));
 			}
 
-			// pull the field traits from the data dictionary class and apply to this field
-			Map<String,Object> masterTraits = mergeFieldTraits(masterField, masterCube, axisName, scope, traits);
+			Map<String,Object> coord = new CaseInsensitiveMap<>(scope);
+			coord.put(axisName,masterField);
```
jdereg: opportunity for groovy-map short hand

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -229,72 +232,45 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+			List<String> traitNames = getTraitNamesForCube(masterCube, (String) scope.get(EFFECTIVE_VERSION_SCOPE_KEY));
+			loadTraitsForField(masterCube, traitNames, traits, coord);
 
-			// check for extended definitions
-			//MODIFIED: Change to use local R_EXTENDS
-			String extension = (String) masterTraits.get(R_EXTENDS);
-			if (!StringUtilities.isEmpty(extension)) {
-				defsToProcess.add(extension);
+			if (traits.containsKey(R_EXISTS) && traits.get(R_EXISTS) ==  null){
```
jdereg: opportunity for groovy-map short hand

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -229,72 +232,45 @@ public class VisualizerHelper extends NCubeGroovyController {
...
-		{
-			String traitName = cell.getKey();
-			Object traitValue = cell.getValue();
-
-			if (!traits.containsKey(traitName))
-			{
-				traits.put(traitName, traitValue);
+			// check for extended definitions
+			if (traitNames.contains(R_EXTENDS)) {
+				coord.put(TRAIT_AXIS, R_EXTENDS);
```
jdereg: opportunity for groovy-map short hand

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -229,72 +232,45 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+				String extension = (String) masterCube.getCell(coord, new HashMap(), NOT_DEFINED);
```
jdereg: [:] instead of new HashMap()

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -348,69 +316,565 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+		private PRIMITIVE_TYPE(Class<?> classType) {
+			this.classType = classType;
+		}
+
+		public Class<?> getClassType() {
+			return this.classType;
+		}
+
+		public static PRIMITIVE_TYPE fromName(String typeName) {
+			for (PRIMITIVE_TYPE type : values()) {
```
jdereg: There is more efficient way to write this.  Use a CaseInsensitiveMap for look up is much case than walking all items in a List.

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -348,69 +316,565 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+	/**
+	 * Load trait values for a given field into the fieldTraits map, ignoring traits already loaded
+	 */
+	private static void loadTraitsForField(NCube classCube, List<String> traitNames, Map<String, Object> fieldTraits, Map<String, Object> coord) {
+		for (String traitName : traitNames) {
+			if (fieldTraits.containsKey(traitName) || R_EXTENDS.equals(traitName)) {
+				continue;
+			}
+
+			coord.put(TRAIT_AXIS,traitName);
```
jdereg: opportunity for groovy-map short hand

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -348,69 +316,565 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+			Object val = classCube.getCell(coord, new HashMap(), NOT_DEFINED);
+			if (hasValue(val)) {
+				fieldTraits.put(traitName, val);
```
jdereg: opportunity for groovy-map short hand

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -348,69 +316,565 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+	 * Determines initial list of fields by extracting column values from axis
+	 */
+	private static void populateAllFieldsFromAxis(Map<String, Map<String, Object>> fieldAndTraits, String axisName, NCube classCube) {
+		for (Column c:classCube.getAxis(axisName).getColumns()) {
+			String fieldName = c.getValueThatMatches().toString();
+
+			Map<String,Object> fieldTraits = fieldAndTraits.get(fieldName);
+			if (fieldTraits==null) {
+				fieldTraits = new LinkedHashMap<>();
+				fieldAndTraits.put(fieldName,fieldTraits);
```
jdereg: opportunity for groovy-map short hand

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -348,69 +316,565 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+		}
+	}
+
+
+	private boolean isFieldValidSince(Map<String, Object> traits, String sourceVersion) {
+		if (!traits.containsKey(R_SINCE)) {
+			return true;
+		}
+
+		Object sinceVersionString = traits.get(R_SINCE);
```
jdereg: opportunity for groovy-map short hand

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -348,69 +316,565 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+		ComparableVersion version = new ComparableVersion(sourceVersion);
+		return version.compareTo(sinceVersion) >= 0;
+	}
+
+	private boolean isFieldValidObsolete(Map<String, Object> traits, String sourceVersion) {
+		if (!traits.containsKey(R_OBSOLETE)) {
+			return true;
+		}
 
+		Object obsoleteVersionString = traits.get(R_OBSOLETE);
```
jdereg: opportunity for groovy-map short hand

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -348,69 +316,565 @@ public class VisualizerHelper extends NCubeGroovyController {
...
 
 	/**
-	 * Get only the traits for the passed in field that have an actual value set on them (containsCell() == true).
-	 * This greatly reduces the number of entries in the returned trait map.
-	 * COPIED: Copied from Dynamis
+	 * Return List of Strings, containing names of trait columns defined on the NCube specified
 	 */
-	private static Map<String, Object> getTraitsForEntry(NCube ncube, String fieldName, String key, Map scope)
+	private static List<String> getTraitNamesForCube(NCube classCube, String sourceVersion) {
+		Axis traitAxis = classCube.getAxis(TRAIT_AXIS);
```
jdereg: Intersection two (2) Sets would be faster than walking both lists.

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -348,69 +316,565 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+	 * Bulk loads value of r:exists for all fields defined
+	 */
+	private void populateExistsTrait(String className, String axisName, Map<String, Object> scope, Map<String, Map<String, Object>> fieldAndTraits, NCube classCube, Map output) {
+		Axis traitAxis = classCube.getAxis(TRAIT_AXIS);
+		if (traitAxis==null || traitAxis.findColumn(R_EXISTS)==null) {
+			return;
+		}
+
+		Map<String, Object> coord = new HashMap<>(scope);
+		coord.put(TRAIT_AXIS, R_EXISTS);
```
jdereg: opportunity for groovy-map short hand

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -348,69 +316,565 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+		for (Column c : classCube.getAxis(axisName).getColumns()) {
+			String fieldName = (String) c.getValue();
+
+			Map<String, Object> fieldTraits = fieldAndTraits.get(fieldName);
```
jdereg: opportunity for groovy-map short hand

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -348,69 +316,565 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+			if (!fieldTraits.containsKey(R_EXISTS)) {
+				coord.put(axisName,fieldName);
```
jdereg: opportunity for groovy-map short hand

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -348,69 +316,565 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+				Boolean exists = getExistsValue(fieldName, className, classCube.getCell(coord,output,NOT_DEFINED));
+
+				if (exists!=null) {
+					fieldTraits.put(R_EXISTS, exists);
```
jdereg: opportunity for groovy-map short hand

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -348,69 +316,565 @@ public class VisualizerHelper extends NCubeGroovyController {
...
 		}
-		return traits;             // returns 1D (slice) of the ncube
 	}
 
+	/**
+	 * Returns the value of the r:exists trait, or null if not set
+	 * @return the boolean value of r:exists, if it exists; otherwise, null
+	 */
+	private static Boolean getExistsValue(String fieldName, String className, Object exists) {
+		if (CLASS_TRAITS.equals(fieldName)) {
```
jdereg: if (CLASS_TRAITS == fieldName) ...

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -348,69 +316,565 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+			return true;
+		}
+
+		if (exists==null)
+		{
+			throw new IllegalStateException(R_EXISTS + EXISTS_TRAIT_CONTAINS_NULL_VALUE + "field: "+ fieldName + ", rpmClass: "+ className);
```
jdereg: Use double " string

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -348,69 +316,565 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+			return null;
+		}
+
+		if (exists instanceof String)
+		{
+			exists = Boolean.valueOf((String)exists);
+		}
+		else if(!(exists instanceof Boolean))
+		{
+			throw new IllegalStateException(R_EXISTS + " must be boolean or string. field: "+ fieldName + ", rpmClass: "+ className);
```
jdereg: double " string

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -348,69 +316,565 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+
+			boolean isNull();
 		}
 
-		public Class<?> getClassType() {
-			return this.classType;
+		/**
+		 * Represents a numeric item in the version item list.
+		 */
+		private static class IntegerItem
```
jdereg: This class looks highly suspect - I believe it may not be needed.

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -348,69 +316,565 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+				implements Item {
+			private static final String[] QUALIFIERS = ['alpha', 'beta', 'milestone', 'rc', 'snapshot', '', 'sp' ];
+
+			@SuppressWarnings("checkstyle:constantname")
+			private static final List<String> _QUALIFIERS = Arrays.asList(QUALIFIERS);
+
+			private static final Properties ALIASES = new Properties();
+			static
+			{
+				ALIASES.put("ga", "");
```
jdereg: opportunity for groovy-map short hand

jdereg: Why not use a Map instead of a Properties (ancient) class?

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -348,69 +316,565 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+			ListItem list = items;
+
+			Stack<Item> stack = new Stack<>();
+			stack.push(list);
+
+			boolean isDigit = false;
+
+			int startIndex = 0;
+
+			for (int i = 0; i < version.length(); i++) {
```
jdereg: should not call .length() on every loop iteration (yes, the method will get call for each iteration).

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -348,69 +316,565 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+				}
+				return buffer.toString();
+			}
+		}
+
+		public ComparableVersion(String version) {
+			parseVersion(version);
+		}
+
+		public final void parseVersion(String version) {
```
jdereg: This whole method looks highly suspect and something should be available already.

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
bheekin: John, I agree with all your recommendations for changes in the VisualizerHelper, but keep in mind that the code in this class is almost exclusively (except for one method, marked ORIGINAL) taken straight out of NCubeRpmClassFactory.java in Dynamis and plopped into here. I have made only minimal modifications in order to make it work outside of Dynamis. Given this and given that we plan to eliminate the whole class once we incorporate Dynamis via Grapes, are you ok if we leave the VisualizerHelper as is?

PR #298, opened 2016-11-19
Visualizer improved handling of initial scope and exceptions

bheekin:
**NOTE!!! Don't merge until the following pull request for n-cube has been merged: https://github.com/jdereg/n-cube/pull/225**

Visualizer changes: 

1. Improved handling of exceptions occurring during load of classes due to missing scope (requires new release of NCube, PR is pending)
2. Added default values to initial minimum scope.
3. Improved handling and messages for missing minimum scope.

### DIFF
--- a/src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
+++ b/src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
@@ -2,6 +2,8 @@ package com.cedarsoftware.util
 
 import com.cedarsoftware.ncube.Column
 import com.cedarsoftware.ncube.NCube
+import com.cedarsoftware.ncube.exception.CoordinateNotFoundException
+import com.cedarsoftware.ncube.exception.InvalidCoordinateException
 import com.google.common.base.Splitter
 import groovy.transform.CompileStatic
 import ncube.grv.method.NCubeGroovyController
@@ -34,7 +36,6 @@ class Visualizer extends NCubeGroovyController
 	public static final String V_MAX = 'v:max'
 
 	public static final String SOURCE_FIELD_NAME = 'sourceFieldName'
-
 	public static final String AXIS_FIELD = 'field'
 	public static final String AXIS_NAME = 'name'
 	public static final String AXIS_TRAIT = 'trait'
@@ -43,15 +44,17 @@ class Visualizer extends NCubeGroovyController
 	public static final String UNSPECIFIED = 'UNSPECIFIED'
 	public static final Map ALL_GROUPS_MAP = [PRODUCT: 'Product', FORM: 'Form', RISK: 'Risk', COVERAGE: 'Coverage', CONTAINER: 'Container', DEDUCTIBLE: 'Deductible', LIMIT: 'Limit', RATE: 'Rate', RATEFACTOR: 'Rate Factor', PREMIUM: 'Premium', PARTY: 'Party', PLACE: 'Place', ROLE: 'Role', ROLEPLAYER: 'Role Player', UNSPECIFIED: 'Unspecified']
 	public static final String[] GROUPS_TO_SHOW_IN_TITLE = ['COVERAGE', 'DEDUCTIBLE', 'LIMIT', 'PREMIUM', 'PRODUCT', 'RATE', 'RATEFACTOR', 'RISK', 'ROLEPLAYER', 'ROLE']
+
+	public static final String EFFECTIVE_VERSION = '_effectiveVersion'
+	public static final String POLICY_CONTROL_DATE = 'policyControlDate'
+	public static final String QUOTE_DATE = 'quoteDate'
 	public static final List DERIVED_SCOPE_KEYS = ['product', 'form', 'risk', 'coverage', 'container', 'deductible', 'limit', 'rate', 'ratefactor', 'premium', 'party', 'place', 'role', 'roleplayer']
 	public static final List DERIVED_SOURCE_SCOPE_KEYS = ['sourceProduct', 'sourceForm', 'sourceRisk', 'sourceCoverage', 'sourceContainer', 'sourceDeductible', 'sourceLimit', 'sourceRate', 'sourceRate Factor', 'sourcePremium', 'sourceParty', 'sourcePlace', 'sourceRole', 'sourceRolepayer']
-	public static final List DEFAULT_AVAILABLE_OPTIONAL_SCOPE_KEYS = ['action', 'businessDivisionCode', 'context', 'Coverage', 'date', 'LocationState', 'policyControlDate', 'quoteDate', 'screen', 'state', 'transaction', 'transactionsubtype']
-	public static final List DEFAULT_AVAILABLE_SCOPE_KEYS = DERIVED_SCOPE_KEYS + DERIVED_SOURCE_SCOPE_KEYS + DEFAULT_AVAILABLE_OPTIONAL_SCOPE_KEYS
+	public static final String SOURCE_SCOPE_KEY_PREFIX = 'source'
 
-	public static final String DEFAULT_SCOPE_VALUE_EFFECTIVE_VERSION = 'X-X-X'
+	public static final List DEFAULT_OPTIONAL_SCOPE_KEYS = ['action', 'businessDivisionCode', 'context', 'Coverage', 'date', 'LocationState', 'screen', 'state', 'transaction', 'transactionsubtype']
+	public static final List DEFAULT_AVAILABLE_SCOPE_KEYS = DERIVED_SCOPE_KEYS + DERIVED_SOURCE_SCOPE_KEYS + DEFAULT_OPTIONAL_SCOPE_KEYS + [POLICY_CONTROL_DATE, QUOTE_DATE, EFFECTIVE_VERSION]
 	public static final String DEFAULT_SCOPE_VALUE = '????'
-	public static final String DEFAULT_SCOPE_VALUE_DATE = 'YYYY-MM-DD'
-	public static final Map DEFAULT_SCOPE = [_effectiveVersion: DEFAULT_SCOPE_VALUE_EFFECTIVE_VERSION, policyControlDate: DEFAULT_SCOPE_VALUE_DATE, quoteDate: DEFAULT_SCOPE_VALUE_DATE]
 	public static final long DEFAULT_LEVEL = 2
 
 	private static final String SPACE = '&nbsp;'
@@ -59,23 +62,23 @@ class Visualizer extends NCubeGroovyController
 	private static final String COMMA_SPACE = ', '
 	private static final String DOUBLE_BREAK = BREAK + BREAK
 
-	public static final String SOURCE_SCOPE_KEY_PREFIX = 'source'
-
 	private VisualizerHelper helper = new VisualizerHelper()
+	static final SafeSimpleDateFormat DATE_TIME_FORMAT = new SafeSimpleDateFormat('yyyy-MM-dd')
 
 	public static final List MANDATORY_RPM_SCOPE_KEYS = [AXIS_FIELD, AXIS_NAME, AXIS_TRAIT]
 	public static final String MISSING_SCOPE = 'missing scope'
 	public static final String UNABLE_TO_LOAD = 'unable to load'
+	public static final String SCOPE_VALUE_NOT_FOUND = 'scope value not found'
 
 	public static final String STATUS_MISSING_START_SCOPE = 'missingStartScope'
-	public static final String NOT_ALL_REQUIRED_SCOPE_KEYS = 'does not contain all of the required scope keys'
 	public static final String STATUS_SUCCESS = 'success'
 
 	private Set messages = []
 	private Set visited = []
-	Set missingRequiredScopeKeys = []
 	Map<String, Set> requiredScopeKeys = [:]
 	Map<String, Set> optionalScopeKeys = [:]
+	String defaultScopeEffectiveVersion
+	String defaultScopeDate
 	Deque<VisualizerRelInfo> stack = new ArrayDeque<>()
 
 	/**
@@ -93,12 +96,15 @@ class Visualizer extends NCubeGroovyController
 	Map buildGraph()
 	{
 		Map options = input.options as Map
+		String cubeName = options.startCubeName as String
 		helper.ncube = ncube
 
 		VisualizerInfo visInfo = new VisualizerInfo()
-		visInfo.startCubeName = options.startCubeName as String
+		visInfo.startCubeName = cubeName
 		Map scope = options.scope as LinkedHashMap
-		visInfo.scope = scope == null ? new LinkedHashMap(DEFAULT_SCOPE) : scope
+		defaultScopeEffectiveVersion = applicationID.version.replace('.', '-')
+		defaultScopeDate = DATE_TIME_FORMAT.format(new Date())
+		visInfo.scope = scope ?: null ? getDefaultScope(cubeName) : scope
 		visInfo.allGroups = ALL_GROUPS_MAP
 		visInfo.availableGroupsAllLevels = []
 		visInfo.groupSuffix = _ENUM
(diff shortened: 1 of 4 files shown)


src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -510,19 +507,20 @@ class Visualizer extends NCubeGroovyController
 		List<String> fields = new ArrayList(traitMaps.keySet())
 		fields.remove(CLASS_TRAITS)
 		String fieldString = "<strong>fields = </strong>"
-		return fieldString + getFieldDetails(fields, SPACE)
+		return fieldString + getFieldDetails(traitMaps, SPACE)
```
jdereg: use "b" instead of "strong"

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -510,19 +507,20 @@ class Visualizer extends NCubeGroovyController
...
 	{
 		StringBuilder fieldDetails = new StringBuilder()
 		fieldDetails.append('')
 
-		fields.each() {
-			fieldDetails.append(BREAK + spaces + it + SPACE)
+		traitMaps.each { k, v ->
+			String fieldName = k as String
+			if (CLASS_TRAITS != fieldName) {
+				fieldDetails.append(BREAK + spaces + k + SPACE)
```
jdereg: use .append() for each addition, dont use + on StringBuilders.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -600,38 +598,65 @@ class Visualizer extends NCubeGroovyController
...
-			if (!scope) {
-				visInfo.scope = new LinkedHashMap(DEFAULT_SCOPE)
-			}
-			String type = Splitter.on('.').split(startCubeName.replace(RPM_CLASS_DOT, '')).first().toLowerCase()
-			visInfo.scope[type] = DEFAULT_SCOPE_VALUE
-			missingRequiredScopeKeys.addAll(visInfo.scope.keySet())
-			return true
-		}
+		String messageSuffix = 'Its default value may be changed as desired.'
+		String messageSuffixType = 'Please replace ' + DEFAULT_SCOPE_VALUE + ' for ' + type + ' with an actual scope value.'
```
jdereg: Use double " string and inline values ${var} instead of string plus (+)

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -600,38 +598,65 @@ class Visualizer extends NCubeGroovyController
...
 
-		if (!scope._effectiveVersion || scope._effectiveVersion == DEFAULT_SCOPE_VALUE_EFFECTIVE_VERSION)
-		{
-			visInfo.scope._effectiveVersion = DEFAULT_SCOPE_VALUE_EFFECTIVE_VERSION
-			missingRequiredScopeKeys << '_effectiveVersion'
+		if (scope) {
+			missingScope = validateScope(visInfo, EFFECTIVE_VERSION, defaultScopeEffectiveVersion, messageSuffix) ? true : missingScope
```
jdereg: These can be written more succinctly as:
missingScope |= validateScope(visInfo, EFFECTIVE_VERSION, defaultScope, messageSuffix)
missingScope |= 2nd
missingScope |= 3rd
This is easier to read and less code.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -600,38 +598,65 @@ class Visualizer extends NCubeGroovyController
...
+			missingScope = validateScope(visInfo, type, DEFAULT_SCOPE_VALUE, messageSuffixType) ? true : missingScope
 		}
-		if (!scope.policyControlDate || scope.policyControlDate == DEFAULT_SCOPE_VALUE_DATE)
-		{
-			visInfo.scope.policyControlDate = DEFAULT_SCOPE_VALUE_DATE
-			missingRequiredScopeKeys << 'policyControlDate'
+		else{
+			missingScope = true
+			visInfo.scope = getDefaultScope(cubeName)
+			messages << 'The scope for the following scope keys was added since it is required: ' +
```
jdereg: Use double quote " string with inline ${var} instead of single quote ' string.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -600,38 +598,65 @@ class Visualizer extends NCubeGroovyController
...
+			}
+		}
+		else{
+			visInfo.scope[key] = value
+			missingScope = true
 		}
 
-		return missingRequiredScopeKeys
+		if (missingScope){
+			messages << 'Scope is required for ' + key + '. ' + messageSuffix
```
jdereg: use double quoted string " with inline ${vars} instead of single quote with string +

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -662,82 +687,113 @@ class Visualizer extends NCubeGroovyController
...
-			rootCause = cause;
-			cause = cause.getCause();
+	private void handleCoordinateNotFoundException(CoordinateNotFoundException e, VisualizerRelInfo relInfo)
+	{
+		String cubeName = e.cubeName
+		String axisName = e.axisName
+		Object value = e.value ?: 'null'
+
+		if (cubeName && axisName) {
+			String message = 'The scope value ' + value + ' for scope key ' + axisName + ' cannot be found on axis ' + axisName + ' on ' + cubeName + getSourceMessage(relInfo) + '.' +
```
jdereg: use double " with inline ${vars} instead of String plus (+) and single quote.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -662,82 +687,113 @@ class Visualizer extends NCubeGroovyController
...
-		return rootCause;
 	}
 
-	private static Set<String> findMissingScope(Map scope, String requiredScopeKeyString)
+	private static String getSourceMessage(VisualizerRelInfo relInfo)
+	{
+		if (relInfo.sourceTraitMaps)
+		{
+			String sourceScopedName = getScopedName(relInfo.sourceTraitMaps)
+			return sourceScopedName ? ', the target of ' + sourceScopedName + ' on ' + relInfo.sourceCube.name : ''
```
jdereg: use double " and inline ${var} instead of single quote and +

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -662,82 +687,113 @@ class Visualizer extends NCubeGroovyController
...
+		Set<String> missingScope = findMissingScope(relInfo.targetScope, requiredKeys)
+		if (missingScope)
+		{
+			Map expandedScope = new LinkedHashMap(visInfo.scope)
+			missingScope.each { String key ->
+				expandedScope[key] = DEFAULT_SCOPE_VALUE
+			}
+			visInfo.scope = expandedScope
+			relInfo.targetTraitMaps = [(CLASS_TRAITS): [(R_SCOPED_NAME): MISSING_SCOPE]]
+			String message = 'Additional scope is required to load ' + cubeName + getSourceMessage(relInfo) +
```
jdereg: use double " and inline ${var} instead of string +

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -662,82 +687,113 @@ class Visualizer extends NCubeGroovyController
...
+					'. Please add scope value(s) for the following scope key(s): ' + String.join(COMMA_SPACE, missingScope) + '.'
+			relInfo.notes << message
+			messages << message
+		}
+		else
+		{
+			throw new IllegalStateException('An InvalidCoordinateException was thrown, but no missing scope keys found for ' + relInfo.targetCube.name + ' and scope ' + visInfo.scope.toString() + '.', e)
```
jdereg: use double " and inline ${vars} instead of String +

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -662,82 +687,113 @@ class Visualizer extends NCubeGroovyController
...
 				missingScope << key
 			}
 		}
 		return missingScope.size() > 0 ? missingScope : null
 	}
 
-	private List getScopeAllReferenced(NCube cube, List<Set> scopeKeys)	{
+	private void handleException(Throwable e, VisualizerRelInfo relInfo)
+	{
+		String message = 'An exception was thrown while loading fields and traits for ' + relInfo.targetCube.name + getSourceMessage(relInfo) + '.' + BREAK + BREAK + '<strong>Message:</strong> ' + BREAK + BREAK + e.message + BREAK + BREAK +  '<strong>Root cause: </strong>'+ BREAK + BREAK + getDeepestException(e).toString()
```
jdereg: use double " and inline ${vars} instead of String +

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -758,4 +814,10 @@ class Visualizer extends NCubeGroovyController
 		requiredScope.remove(AXIS_TRAIT);
 		return requiredScope;
 	}
+
+	private static String getTypeFromCubeName(String cubeName)
+	{
+		return Splitter.on('.').split(cubeName.replace(RPM_CLASS_DOT, '')).first().toLowerCase()
```
jdereg: Can this just be (using the minus operator with String):
String name = "rpm.class.FooBar"
String className = name - "rpm.class."
println className

Result: FooBar

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -74,11 +74,25 @@ public class VisualizerHelper extends NCubeGroovyController {
...
-
+		removeNotExistsFields(traitMaps)
 		return traitMaps
 	}
 
 	/**
+	 * ORIGINAL: Not copied from Dynamis
+	 */
+	private static void removeNotExistsFields(Map<String, Map<String, Object>> traitMaps) {
+		List notExistFields = []
```
jdereg: This code passes through the field list twice to remove the fields where not exists is false.  This can be done in one (1) pass using an iterator.  Iterator i; ... i.remove() when you need to remove the item at the current iterator spot.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -391,8 +401,8 @@ class Visualizer extends NCubeGroovyController
 		String sourceCubeEffectiveName = getEffectiveName(sourceCube, sourceTraitMaps)
 		String targetCubeEffectiveName = getEffectiveName(targetCube, relInfo.targetTraitMaps)
 		edgeMap.id = String.valueOf(relInfo.id)
-		edgeMap.from = sourceCube.name + '_' + relInfo.sourceScope.toString()
-		edgeMap.to = targetCube.name + '_' + relInfo.targetScope.toString()
+		edgeMap.from = sourceCube.name + '_' + relInfo.sourceScope.toString()  //Note: Cannot use GString here since vis.js can't handle it
```
jdereg: You have to call .toString() on a GString before it can be used on the JS side. 
\map.key = "This needs to be converted to regular string if going to js".toString()

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -450,7 +460,7 @@ class Visualizer extends NCubeGroovyController
 		sb.append(line)
 
 		Map nodeMap = [:]
-		nodeMap.id = targetCube.name + '_' + targetScope.toString()
+		nodeMap.id = targetCube.name + '_' + targetScope.toString() //Note: Cannot use GString here since vis.js can't handle it
```
jdereg: call .toString() on GString

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -600,38 +634,71 @@ class Visualizer extends NCubeGroovyController
...
+		boolean missingScope
+		String cubeName = visInfo.startCubeName
 		Map scope = visInfo.scope
+		String type = getTypeFromCubeName(cubeName)
+
+		String messageSuffix = "Its default value may be changed as desired."
+		String messageSuffixType = "Please replace ${DEFAULT_SCOPE_VALUE} for ${type} with an actual scope value."
+
+		if (scope) {
+			missingScope = validateScope(visInfo, EFFECTIVE_VERSION, defaultScopeEffectiveVersion, messageSuffix) ? true : missingScope
```
jdereg: These should be OR-EQUALS (  |=   )

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -758,4 +892,10 @@ class Visualizer extends NCubeGroovyController
 		requiredScope.remove(AXIS_TRAIT);
 		return requiredScope;
 	}
+
+	private static String getTypeFromCubeName(String cubeName)
+	{
+		return (cubeName - RPM_CLASS_DOT).toLowerCase()
```
jdereg: Any reason we need this in lower case?  That is a very expensive call to make.  If this value is being placed into a CaseInsensitiveSet or Map, then it does not matter.  I cant do search usages from GitHub, so hard for me to check at the moment.

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -74,11 +74,25 @@ public class VisualizerHelper extends NCubeGroovyController {
...
+		traitMaps.keySet().each{ String fieldName->
```
jdereg: Instead of looping through the keySet() twice, use an iterator:

Iterator i = traitMaps.keySet().iterator()
while (i.hasNext)
{
    String key = i.next()
    if (field should not be there)
       i.remove()
}

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
bheekin: Aha, that makes sense. It bit me quite hard before I realized why the visualizer was blowing up on the client.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
bheekin: I'm using that method in several places now (pending change) and in some it's used as part of an error message to the user. I will change it to not always to .toLower. Instead, I will do that in the select places where I need it.

src/main/groovy/com/cedarsoftware/util/VisualizerHelper.groovy
```diff
@@ -74,11 +74,28 @@ public class VisualizerHelper extends NCubeGroovyController {
 		} else{
 			loadRpmClassFields(RPM_CLASS, cube.name.split(RPM_CLASS_DOT)[1], scope, traitMaps, [:])
 		}
-
+		removeNotExistsFields(traitMaps)
 		return traitMaps
 	}
 
 	/**
+	 * ORIGINAL: Not copied from Dynamis
```
jdereg: Thanks for making this change - it will help performance.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -600,38 +634,72 @@ class Visualizer extends NCubeGroovyController
...
-		String startCubeName = visInfo.startCubeName
+		boolean missingScope
+		String cubeName = visInfo.startCubeName
 		Map scope = visInfo.scope
+		String type = getTypeFromCubeName(cubeName).toLowerCase()
+
+		String messageSuffix = "Its default value may be changed as desired."
+		String messageSuffixType = "Please replace ${DEFAULT_SCOPE_VALUE} for ${type} with an actual scope value."
+
+		if (scope) {
```
jdereg: Is groovy giving you an error with the OR-Equal?

validateScope() should return boolean instead of Boolean, unless you are allowing for null?

bheekin: Changing to return boolean instead of Boolean did the trick! Not sure why I was using Boolean... I think it was left over from playing around with other ways of handling the logic. Will push this change up in just a bit.

bheekin: I spoke to soon on the |= operator working for missing scope... It’s not working with my logic. Not sure what I was doing/seeing to think it did work since my unit tests (that I have over in CUBE for now) are failing. I have reverted back to using the Elvis operator as you can see in the PR now.

I also renamed the validateScope method to addMissingScope and renamed the variable missingScope to hasMissincScope. 

The purpose of the (now named) addMissingScope method is to add any minimum scope that is missing, then return true if something was missing, else return false. I also only want the (now named) hasMissingScope variable to change from its initial value of false if addMissingScope returns true. That is, if the first call to addMissingScope changes  hasMissingScope to true, I don’t want the second call to turn it back to false if the second call returns false. Thus the Elvis operator.

PR #301, opened 2016-11-23
A few more cleanup items for the Visualizer

bheekin:
Changed .compareTo to > and <.
Extracted code relating to building the node label into a separate method for readability.
Removed TODO’s about .lowerCase in getScopeRelativeToSource() and getDefaultScope(). Need to convert to lower case since the scope in both cases is displayed to the user.

### DIFF
--- a/src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
+++ b/src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
@@ -62,8 +62,8 @@ class Visualizer extends NCubeGroovyController
 	private static final String COMMA_SPACE = ', '
 	private static final String DOUBLE_BREAK = "${BREAK}${BREAK}"
 
-	private VisualizerHelper helper = new VisualizerHelper()
-	static final SafeSimpleDateFormat DATE_TIME_FORMAT = new SafeSimpleDateFormat('yyyy-MM-dd')
+	private static final int NODE_LABEL_MAX_LINE_LENGTH = 16
+	private static final BigDecimal NODE_LABEL_LINE_LENGTH_MULTIPLIER = 1.2
 
 	public static final List MANDATORY_RPM_SCOPE_KEYS = [AXIS_FIELD, AXIS_NAME, AXIS_TRAIT]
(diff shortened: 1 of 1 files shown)


src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -62,8 +62,8 @@ class Visualizer extends NCubeGroovyController
 	private static final String COMMA_SPACE = ', '
 	private static final String DOUBLE_BREAK = "${BREAK}${BREAK}"
 
-	private VisualizerHelper helper = new VisualizerHelper()
-	static final SafeSimpleDateFormat DATE_TIME_FORMAT = new SafeSimpleDateFormat('yyyy-MM-dd')
+	private static final int NODE_LABEL_MAX_LINE_LENGTH = 16
+	private static final BigDecimal NODE_LABEL_LINE_LENGTH_MULTIPLIER = 1.2
```
jdereg: Can this be a double instead of a BigDecimal?  BigDecimal is the slowest possible mathematical primitive to use - it's accurate but much slower than double.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -428,30 +428,42 @@ it cannot be loaded as an rpm.class in the visualization."""
...
 			sb.append(label)
 			sb.append('\n')
-			sb.append('-' * Math.floor(len * 1.2))
+			sb.append('-' * Math.floor(len * NODE_LABEL_LINE_LENGTH_MULTIPLIER))
 			sb.append('\n')
 		}
 
-        // TODO: This code needs to be broken into named methods that explain what it is doing
-		String labelName = getDotSuffix(getEffectiveName(targetCube, targetTraitMaps))
+		String labelName = getDotSuffix(getEffectiveName(relInfo.targetCube, relInfo.targetTraitMaps))
```
jdereg: can we get a new name for the method 'getDotSuffix()' ? I know you may not have written it, but what is it really fetching?

PR #303, opened 2016-11-24
Added load of available scope value for scope keys + other minor fixes

bheekin:
1. Changed ALL_GROUPS_MAP.keySet() to be a static constant.
2. Replaced one last occurrence of columns.find with axis.findColumn.
3. Added load of available scope values for scope keys from both EPM cubes and from ENT_APP. This will be made available to UI so that the scope picker can have dropdowns with values. For now, it’s used to display available scope values to users when they get toasts about missing scope. Some of the new code will be refactored to be moved to n-cube (TODO).
4. Moved messy GString messages into their own methods for readability.
5. Extracted local variable for source field trait map.
6. Changed label line length multiplier to be declared as double.

### DIFF
--- a/src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
+++ b/src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
@@ -1,7 +1,10 @@
 package com.cedarsoftware.util
 
-import com.cedarsoftware.ncube.Column
+import com.cedarsoftware.ncube.ApplicationID
+import com.cedarsoftware.ncube.Axis
 import com.cedarsoftware.ncube.NCube
+import com.cedarsoftware.ncube.NCubeManager
+import com.cedarsoftware.ncube.ReleaseStatus
 import com.cedarsoftware.ncube.RuleInfo
 import com.cedarsoftware.ncube.exception.CoordinateNotFoundException
 import com.cedarsoftware.ncube.exception.InvalidCoordinateException
@@ -24,6 +27,7 @@ class Visualizer extends NCubeGroovyController
 	public static final String RPM_ENUM_DOT = 'rpm.enum.'
 	public static final String CLASS_TRAITS = 'CLASS_TRAITS'
 	public static final String DOT_CLASS_TRAITS = '.classTraits'
+	public static final String DOT_TRAITS = '.traits'
 	public static final String R_RPM_TYPE = 'r:rpmType'
 	public static final String R_EXISTS = 'r:exists'
 	public static final String V_ENUM = 'v:enum'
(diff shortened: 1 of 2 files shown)


src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -892,4 +891,146 @@ ${DOUBLE_BREAK}${t.toString()}${DOUBLE_BREAK}<b>Stack trace: </b>${DOUBLE_BREAK}
...
+	}
+
+	private Map loadAvailableScopeValues()
+	{
+		Map valuesByKey = new CaseInsensitiveMap()
+
+		//Values for Risk, SourceRisk, Coverage, SourceCoverage, etc.
+		DERIVED_SCOPE_KEYS.each{ key ->
+			String cubeName = RPM_SCOPE_CLASS_DOT + key + DOT_TRAITS
+			List values = getCube(cubeName).getAxis(key).columnsWithoutDefault.collect{it.value}
```
jdereg: This code could call the new getColumnValues() method below.
List values = getColumnValues(ncube.getApplicationID(), ncube.name, axisName)

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -892,4 +891,146 @@ ${DOUBLE_BREAK}${t.toString()}${DOUBLE_BREAK}<b>Stack trace: </b>${DOUBLE_BREAK}
 	{
 		return (cubeName - RPM_CLASS_DOT)
 	}
+
+	private Set loadAvailableScopeValues(VisualizerInfo visInfo, String cubeName, String key)
+	{
+		Axis axis = getCube(cubeName).getAxis(key)
```
jdereg: This code could call the new getColumnValues() method below.
List values = getColumnValues(ncube.getApplicationID(), ncube.name, axisName)
You may need to make the new method below have the same check for axis name not being found.  Either way, make the method below more general so all places can use it.  No need to repeat this logic.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -892,4 +891,146 @@ ${DOUBLE_BREAK}${t.toString()}${DOUBLE_BREAK}<b>Stack trace: </b>${DOUBLE_BREAK}
...
+		valuesByKey[BUSINESS_DIVISION_CODE] = getColumnValues(entAppAppId, BUSINESS_DIVISION_CUBE_NAME, BUSINESS_DIVISION_CODE)
+		List stateValues = getColumnValues(entAppAppId, STATE_CUBE_NAME, STATE)
+		valuesByKey[STATE] = stateValues
+		valuesByKey[LOCATION_STATE] = stateValues
+
+		return valuesByKey
+	}
+
+	private static List getColumnValues(ApplicationID applicationID, String cubeName, String axis)
+	{
```
jdereg: If this method is changed slightly to do the check to see if the Axis is there, and return an empty List if not, otherwise return the column values, then all the other places in the code could use this method.

Also, n-cube by design, enforces that the Axis contain unique values.  Because of this, this method should return a Set, not a List.

PR #309, opened 2016-12-01
Visualizer: Changes to handling of scope in js

bheekin:
1. Scope is now saved off to local storage and available when re-opening the cube for visualization or using the back button to navigate back to a visualization.

2. Fixed issues with the scope picker. It is now displaying already picked scope keys and is able to add scope keys.
            TODO: The key in the scope picker is case sensitive, which doesn’t play well with the case insensitive scope that comes across from the server (Product vs. product, etc.).
            TODO: When a key is picked in the scope picker, make the value a drop-down populated with available scope values.

3. Added temporary code to handle cube names with "_" instead of "." (e.g. rpm_class_product instead of rpm.class.product) in nce.getSelectedCubeName()after a page refresh (not always, but often).

            TODO: Figure out what the deal is here and find a permanent solution.

### DIFF
--- a/src/main/webapp/js/visualize.js
+++ b/src/main/webapp/js/visualize.js
@@ -33,7 +33,7 @@ var Visualizer = (function ($) {
     var _edges = null;
     var _networkData = null;
     var _scope = null;
-    var _keepScope = false;
+    var _keepCurrentScope = false;
     var _availableScopeKeys = [];
     var _selectedGroups = null;
     var _availableGroupsAtLevel = null;
@@ -43,7 +43,7 @@ var Visualizer = (function ($) {
     var _maxLevel = null;
     var _nodeCount = null;
     var _groupSuffix = null;
-    var _startCube = null;
+    var _selectedCubeName = null;
     var _hierarchical = false;
     var _loadTraits = false;
     var _selectedLevel = null;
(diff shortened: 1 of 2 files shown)


src/main/webapp/js/visualize.js
```diff
@@ -997,31 +1016,38 @@ var Visualizer = (function ($) {
...
+    //TODO: occasionally contains a cube name with "_" instead of "." (e.g. rpm_class_product instead of
+    //TODO: rpm.class.product) after a page refresh.
+    function getStorageKey(nce, prefix) {
+        //return prefix + ':' + nce.getSelectedTabAppId().app.toLowerCase() + ':' + nce.getSelectedCubeName().toLowerCase();
+        return prefix + ':' + nce.getSelectedTabAppId().app.toLowerCase() + ':' + _selectedCubeName.toLowerCase();
+    }
+
     function saveScope() {
-        saveOrDeleteValue(_savedScope, getStorageKey(_nce, SCOPE_MAP));
+        saveOrDeleteValue(_scope, getStorageKey(_nce, SCOPE_MAP));
```
jdereg: The scope key for local storage should include APP_ID and N-Cube name.

bheekin: The function getStorageKey in index.js does the following, so I think we are ok: 

 function getStorageKey(nce, prefix) {
        return prefix + ':' + nce.getSelectedTabAppId().app.toLowerCase() + ':' + nce.getSelectedCubeName().toLowerCase();
  }

PR #322, opened 2016-12-12
Visualizer: Add wait messages and more

bheekin:
1.	Refined the network click event to distinguish better between single click (show info panel) and double-click (cluster the node)
2.	Added “cluster” to label on clusters (double clicked nodes).
3.	Added network status and iterations to stabilize network to Einstein section. The latter parm (iterations to stabilize) might help us get ‘em boys in from recess.
4.	Saving Einstein setting to local storage.
5.	Fixed bug that caused the “all” and “none” listeners for groups to multiply. Should be faster now to hit all or none for groups.
6.	Added toasts for “Loading visualizer”, “Loading traits”, “Updating network” and “Stabilizing network”.
7.	Fixed issue with old toast not clearing when reloading a different visualizer tab or with different scope.


src/main/webapp/js/index.js
```diff
@@ -4390,6 +4391,11 @@ var NCE = (function ($) {
             _errorId = null;
         }
     }
+
```
jdereg: This 
(clearAllErrors()) seems like it should not be needed.  For one, the internal frames should not know about trigger - tjhey should interact with an API that index (nce) provides to them.  This locks the implementation of the toast now into the iFrame.

PR #323, opened 2016-12-12
Visualizer: Change clearAllErrors function to use Gritter's API method removeAll.

bheekin:
Change new clearAllErrors function to use Gritter's API method removeAll. Also, allow user to double click missing scope key in scope input text.


jdereg:
Make sure you import the 'Inspections' for Intellij.  When I look at Visualize.js, I see nearly 100 inspection issues, including many missing semi-colons which can trip up browsers.  Groovy of course allows, this but in Javascript, it barely supports missing semi-colons, so our standard is to always use them.  I highly recommend using the 'standard' inspections imported into Intellij, and then look through the .js code and make it's appropriate suggested fixes.

PR #326, opened 2016-12-14
Add default value support on an Axis for the default column.

tympollack:


src/main/webapp/js/ntwobe.js
```diff
@@ -2210,16 +2221,23 @@ var NCubeEditor2 = (function ($) {
     }
 
     function getColumnDefault(column) {
-        var val;
+        var val, axis;
         if (column.hasOwnProperty(METAPROPERTIES.DEFAULT_VALUE)) {
             val = column[METAPROPERTIES.DEFAULT_VALUE];
+        } else if (column.value === DEFAULT_TEXT) {
```
jdereg: The column.value should be null for default columns.  Not sure why it isn't that way here.  We should not be checking for the String 'default' to determine if a column is the default.  The value being null (only allowed on default) is the correct check for a default column.

tympollack: On the front end, when we set up the cube's data we create a column on the axis with the correct default id for the axis and a value of 'Default' in order to display in the table. All of this is for display purposes only, it does not affect data in the cube.

PR #327, opened 2016-12-15
Visualizer: Add onclick event to toasts. 

bheekin:
1.	Added click event handler to toasts (aka gritter notes). Also, added onNoteClick method to index.js. This method is called by the handler and turns around and calls a method of the same name on other NCE iFrames. 
2.	Added functionality to visualizer to allow user to click available scope values in toasts and in info panel in order to populate any missing scope.
3.	Changed the following network options parms: physics.minVelocity (from 0.85 to 1), edges.selectionWidth (from default to 2) and edges.smooth.enabled (from true to false). Note, edges smooth.type does not need to change since smooth is now disabled.


src/main/webapp/js/jquery.gritter.js
```diff
@@ -371,6 +380,15 @@
         },
 
         /**
+         * Get notice wrapper
+         * @return e The notice wrapper element.
+         */
+        noticeWrapper: function()
+        {
+            return $('#gritter-notice-wrapper');
```
jdereg: This needs to be stored in a variable otherwise the query runs every time.

PR #337, opened 2016-12-21
Visualizer: Create VisualizerConfig json file + other changes

bheekin:
This is the reworked PR based on code review feedback + with some additional changes. 

1.	Created a new VisualizerConfig json file in NCE that holds configs for the visualizer. Configurations are only loaded first time through or if the app id changes. 
2.	Moved the TypesToAdd cube from UD.REF.APP to a json file in NCE. Also moved typesToAdd method in Visualizer.groovy into VisualizerInfo.groovy.
3.	Refactored to pass the VisualizerInfo object back and forth between the .js and the .groovy. Previously, the object was only passed down to the .js and had to be recreated in .groovy.
4.	No longer passing selectedGroups and selectedLevel back and forth from server from js. Not necessary.
5.	Commented out code relating to gathering availableScopeKeys. Not needed currently, but I will revisit this and may use it again. If not, I’ll remove the code.
6.	Removed left-over code from when we still had the feature to load traits for all nodes.

### DIFF
--- a/src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
+++ b/src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
@@ -30,14 +30,9 @@ class Visualizer
 	 *
 	 * @param applicationID
 	 * @param options - a map containing:
-	 *            String startCubeName, name of the starting cube
-	 *            Map scope, the context for which the visualizer is loaded
-	 *            Set<String> availableScopeKeys, scope keys available in the visualization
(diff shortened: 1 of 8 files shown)


jdereg:
I am merging this as is, however, soon I want to see the 'types to add' code removed for the following reasons.  1) It has no concept of field.  It is relating objects on type-only, which is not valid in many circumstances.  2) This information, rather than being stored as configuration, can be 'discovered.'  And in discovering it, the concept of 'field' can be handled.

PR #350, opened 2017-01-07
Visualizer: Improved handling of values mode for non-rpm cubes

bheekin:
Visualizer: Improved handling of values mode for non-rpm cubes

1.	Refactored loading of cell values for non-rpm. Simplified logic and broke out into a new groovy class, VisualizerCellInfo.
2.	Refactored handling of exceptions during cell execution. The exceptions are logged for each cell for non-rpm cubes. For rpm cubes they are still logged for the class as a whole. Created a new helper class that is shared by non-rpm and rpm cubes for exception handling.
3.	Improved display of coordinates and cell values. Will do more work in this area. At a minimum, sort coordinates and provide show/hide of cell values.
4.	All cells are executed when user clicks “Show values”. Three categories of cells: 
•	 Executed ok - green checkmark 
               Shows non-executed and executed value
•	InvalidCoordinateException or CoordinateNotFoundException – blue question mark
              Shows non-executed value and message with clickable scope options.
•	Any other exception – red exclamation mark
               Shows non-executed value and exception info
5.	Removed option to execute all. No need for it, at least for now. 
6.     Add wrapping in info panel pre tags.
7.     Add show/hide of cell values.
8.     Add expand/collapse all cell values.
9.     List out axes names in info panel for non-rpm.
10.   Sort coordinates displayed in info panel. Doubtful that this sort is the most efficient, have a question out to John to discuss.
11.	Removed “under construction” message. I think it’s in good enough shape to qualify without it.
12.	CubeM in bheekin2 is a good example of the four categories of cells.
13.	CubeI in bheekin2 has a ton of different types of cell values and is another good example.


src/main/groovy/com/cedarsoftware/util/VisualizerRelInfo.groovy
```diff
@@ -59,76 +53,39 @@ class VisualizerRelInfo
...
-	private String getCoordinateString(Map<String, Object> coordinates)
-	{
-		coordinates.each {String key, Object value ->
-			if (!value)
-			{
-				coordinates[key] = 'null'
+			cellInfo.sort { VisualizerCellInfo cellInfo ->
+				cellInfo.coordinate.toString()
 			}
 		}
```
bheekin: I have a feeling there are better and more efficient ways sort the list of visCellInfo objects...  currently doing a .toString() on each visCellInfo's coordinate. Will send a separate email that has an example of the sort goal I'm trying to achieve. Suggestions welcome.

PR #367, opened 2017-01-13
Visualizer: Add unit tests for ncube (non-rpm) visualizer

bheekin:
Added unit tests that I previously had running in cube-resources to NCE so they can run regularly with NCE.  

To get the tests running in NCE, also added a TestingHelper and a TestingNCubePersister. Not sure this setup is what we want long-term, but it gets the job done now. 

Also, I believe the three classes TestingNCubeManagerHelper, TestNCubeController, and TestVisualizer in test/com.cedarsoftware.controller are obsolete (TestVisualizer is for sure), unless there is a longer term vision for them. Same goes for the rpm.xxx json files under test/resources. I can delete as part of this PR if I get the thumbs up.


jdereg:
Before I merge this, I have a couple of suggested changes that I think will help.  First of all, go ahead and make your recommended changes regarding obsolete testing classes.  

The `TestingHelper` and `TestingNCubePersister` are a good idea but can be expanded in their role by renaming them.  `TestingHelper` should be renamed to `NCubeResourceLoader`.  And `TestNCubePersister` should be renamed to the `NCubeResourcePersister`, since it implements the NCubePersister API and load n-cubes resources.  It just happens to be used by tests.

Next, you should look at this link: http://stackoverflow.com/questions/3923129/get-a-list-of-resources-from-classpath-directory . It will explain how to enumerate resources on the classpath.  This will allow the resource persister to be more completely implemented (rather than stub nearly all methods off, most of the ReadOnlyPersister APIs could be implemented, increasing the value of ther ResourcePersister.)

bheekin:
Here's the latest commit based on your recommendations, or rather, based on how I understood your recommendations; let me know if you had something different in mind. If that's the case, perhaps we can chat later today?

Renamed TestNCubePersister to NCubeResourcePersister.
Reworked load cube logic in the persister.
Deleted obsolete test classes and test json.

src/test/groovy/com/cedarsoftware/ncube/NCubeResourcePersister.groovy
```diff
@@ -0,0 +1,204 @@
...
+            NCubeManager.addCube(appId, cube)
+            cubes[resource.filename - JSON_FILE_SUFFIX] = cube
+        }
+        return cubes
+    }
+
+    @Override
+    List<NCubeInfoDto> search(ApplicationID applicationID, String s, String searchValue, Map options)
+    {
+       return []
```
jdereg: I believe this API can be implemented via the code here: http://stackoverflow.com/questions/3923129/get-a-list-of-resources-from-classpath-directory

src/test/groovy/com/cedarsoftware/ncube/NCubeResourcePersister.groovy
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    @Override
+    NCube loadCubeBySha1(ApplicationID applicationID, String s, String s1) {
+        return null
```
jdereg: I believe this API can be implemented via the code here: http://stackoverflow.com/questions/3923129/get-a-list-of-resources-from-classpath-directory . It would scan the resources, load the n-cube, calc their sha-1, and return matching one.

PR #369, opened 2017-01-16
Visualizer: Add tests for rpm visualizer

bheekin:
1. Moved the RpmVisualizerTest into NCE from cube-resources. Along with that go 33 json test files. Test coverage for the visualizer groovy code is now 95%+.
2. Extracted more constants.
3. Removed cache check in NCubeResourcePersister.loadCube().
4. Remove obsolete configs in VisualizerConfig.


bheekin:
Very good point. I was considering that issue and addressed it by altering the rpm cubes to have just a few entities with new names and only a minimal set of traits and fields. BUT, let's follow up on this and make sure we are in the clear.

PR #375, opened 2017-01-19
Bugfix for left pane sizing when using search options.

tympollack:


src/main/webapp/js/index.js
```diff
@@ -253,8 +253,15 @@ var NCE = (function ($) {
                 },PROGRESS_DELAY);
             }
         });
+    }
+    
+    function setupWestSplitter() {
+        var toggleClass, modCubesOpen;
+        modCubesOpen = getModCubesOpen();
+        modCubesOpen = modCubesOpen ? parseInt(modCubesOpen) : 0;
```
cstroliadavis: This could probably be written more simply like:

``` javascript
  modCubesOpen = parseInt(getModCubesOpen() || 0);
```

src/main/webapp/js/index.js
```diff
@@ -314,20 +312,32 @@ var NCE = (function ($) {
...
+
+    function southPanelToggle(open) {
+        var addClass, removeClass;
+        if (open) {
+            addClass = 'glyphicon-collapse-down';
+            removeClass = 'glyphicon-collapse-up';
+        } else {
+            addClass = 'glyphicon-collapse-up';
+            removeClass = 'glyphicon-collapse-down';
+        }
```
cstroliadavis: It seems like something like this would be simpler:
``` javascript
var addClass, removeClass, iconBase;

iconBase = 'glyphicon-collapse-';
addClass = iconBase + (open ? 'down' : 'up');
removeClass = iconBase + (open ? 'up' : 'down');
```

jdereg: The version Tym had is more elegant and easier to change.  I would not introduce all this logic to try to save a little base portion of the String.

PR #376, opened 2017-01-19
Fix diff display for metaproperties

jordankomnick:


src/main/webapp/js/index.js
```diff
@@ -4461,6 +4461,11 @@ var NCE = (function ($) {
         if (delta.type.name === DELTA.TYPE.UPDATE) {
             return getObjectTextArray(isSource ? delta.sourceVal : delta.destVal);
         }
+        if ([DELTA.LOC.AXIS_META, DELTA.LOC.CELL_META, DELTA.LOC.COLUMN_META, DELTA.LOC.NCUBE_META].indexOf(delta.loc.name) > -1) {
+            var sourceInput = delta.sourceVal ? delta.sourceVal : [];
+            var destInput = delta.destVal ? delta.destVal : [];
+            return getObjectTextArray(isSource ? sourceInput : destInput);
+        }
         return getObjectTextArray(isSource ? delta.sourceList : delta.destList);
     }
```
tympollack: The null was mostly getting checked bc certain object values from the DeltaProcessor come over as null if not used (or values like default column or cells with null). This change is less specifically for metaproperties and more for DeltaTypes of ADD and DELETE. We used to display the full list and show what changed between the two vs having one side completely blank for adds and deletes, which are a good portion of changes. This code can be more generalized the following. Also, please remember to declare variables at the top of function scope in js.
```
        var sourceInput, destInput;
        if (delta.type.name === DELTA.TYPE.UPDATE || delta.loc.name === DELTA.LOC.NCUBE) {
            return getObjectTextArray(isSource ? delta.sourceVal : delta.destVal);
        }
        if ([DELTA.TYPE.ADD, DELTA.TYPE.DELETE].indexOf(delta.type.name) > -1) {
            sourceInput = delta.sourceVal ? delta.sourceVal : [];
            destInput = delta.destVal ? delta.destVal : [];
            return getObjectTextArray(isSource ? sourceInput : destInput);
        }
        return getObjectTextArray(isSource ? delta.sourceList : delta.destList);
```

PR #379, opened 2017-01-20
pixel issues

tympollack:


src/main/webapp/js/constants.js
```diff
@@ -35,7 +35,8 @@ var MODAL_TOO_FAR_LEFT = -250;
 var MAIN_SPLITTER_DEFAULTS = {
     TOGGLER_LENGTH_OPEN: 60,
     WEST_SIZE: 250,
-    WEST_MIN_SIZE: 140,
+    WEST_MIN_SIZE: 200,
+    WEST_MAX_SIZE: 700,
```
jdereg: any reason for the max?

tympollack: Just to give a range instead of dragging all the way across the screen.

PR #382, opened 2017-01-23
Temp fix for permissions when unable to commit.

tympollack:
I call this a "temp fix" bc it will go away once we get the PR framework in place. But for right now, since we stop people at commit instead of update, we needed a good way to communicate ahead of time that they can't commit their own changes for certain cubes.

### DIFF
--- a/src/main/webapp/js/constants.js
+++ b/src/main/webapp/js/constants.js
@@ -140,7 +140,8 @@ var SAVED_INFO = {
     NUMBER_OF_FROZEN_COLUMNS: 'numFrozenCols',
     SEARCH_QUERY: 'searchQuery',
     SHOULD_LOAD_ALL_FOR_SEARCH: 'shouldLoadAllForSearch',
-    VIEW_POSITION: 'position'
+    VIEW_POSITION: 'position',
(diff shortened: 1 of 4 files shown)


src/main/webapp/js/constants.js
```diff
@@ -140,7 +140,8 @@ var SAVED_INFO = {
     NUMBER_OF_FROZEN_COLUMNS: 'numFrozenCols',
     SEARCH_QUERY: 'searchQuery',
     SHOULD_LOAD_ALL_FOR_SEARCH: 'shouldLoadAllForSearch',
-    VIEW_POSITION: 'position'
+    VIEW_POSITION: 'position',
+    HAS_BEEN_WARNED_ABOUT_UPDATING_IF_UNABLE_TO_COMMIT_CUBE: 'hasBeenWarnedAboutUpdatingIfUnableToCommitCube'
```
gaigstroliac: Might just be me, but this seems overly long. 

I don't know if there are certain conventions that need to be followed here, but could it be shortened like this:

`WAS_WARNED_ABOUT_UPDATE_IF_CANT_COMMIT_CUBE`

tympollack: It's just temp code anyway. Once we get the PR system in place, everything for this will be removed.

PR #386, opened 2017-01-24
Update handsontable to v0.30.1

tympollack:


jdereg:
It looks like Handson table is now using ZeroClipboard?  Does that mean we can remove our special ctrl-c, ctrl-x, and ctrl-v handling in favor or ZeroClipboard?  It may allow us to have the Cut, Copy, and Paste options on a menu.  Originally, I was unable to do that, because I am relying on the technique that traps the key-event and routing it to a hidden text area.

PR #390, opened 2017-01-28
Visualizer: Code cleanup per code review feedback

bheekin:
1. Removed the addUnboundAxes method in RpmVisualizerHelper and the four calls to it. Instead making sure that output is passed around between all the different cell execute calls that are made during load of fields and traits for a class. This way, the output returned from loading fields and traits has all the unbound axes info needed.
2. Improved the handling of unbound axes in RpmVisualizerRelInfo. Much simplified and should run faster.   
3. Simplified getScopeValues method on VisualizerInfo. It was working, but was quite the mess.
4. Fixed bug in loading of cell values for non-rpm cubes. Was not passing all the scope available to the cube when loading the cell values.
5. Changes per Tym's code review suggestions.
6. Changed references to “key” to have more meaningful name “scopeKey”.


src/main/groovy/com/cedarsoftware/util/RpmVisualizerRelInfo.groovy
```diff
@@ -442,19 +475,19 @@ class RpmVisualizerRelInfo extends VisualizerRelInfo
 	{
 		String cubeName = e.cubeName
 		String effectiveNameByCubeName = effectiveNameByCubeName
-		StringBuilder sb = new StringBuilder()
+		StringBuilder sb = new StringBuilder(ADDITIONAL_SCOPE_REQUIRED_TO_LOAD)
```
jdereg: What is line 477 doing?  It makes no sense to me.

src/main/groovy/com/cedarsoftware/util/VisualizerInfo.groovy
```diff
@@ -33,8 +33,8 @@ class VisualizerInfo
     Set<String> availableGroupsAllLevels
     Set<String> messages
 
-    Map<String, Set<Map<String, Set<Object>>>> optionalScopeValues = [:]
-    Map<String, Set<Map<String, Set<Object>>>> requiredScopeValues = [:]
+    Map<String, Map<String, Set<Object>>> optionalScopeValues = [:]
+    Map<String, Map<String, Set<Object>>> requiredScopeValues = [:]
```
jdereg: Should these be LinkedHashMaps [:] or CaseInsenstiveMaps?

src/main/groovy/com/cedarsoftware/util/VisualizerInfo.groovy
```diff
@@ -141,46 +141,24 @@ class VisualizerInfo
...
-        else
-        {
-            Set<Object> values =  getColumnValues(appId, cubeName, key)
-            cubeMap = [(cubeName): values]
-            valuesByCubeNames << cubeMap
-            scopeValues[key] = valuesByCubeNames
-            return values
-        }
+        //The key to the map scopeValues is a scope key. The map contains a map of scope values by cube name.
+        Map<String, Set<Object>> scopeValuesForScopeKey = scopeValues[scopeKey] as Map ?: [:]
```
jdereg: Should this be a LinkedHashMap [:] or a CaseInsenstiveMap ?

src/main/groovy/com/cedarsoftware/util/RpmVisualizerRelInfo.groovy
```diff
@@ -376,61 +376,94 @@ class RpmVisualizerRelInfo extends VisualizerRelInfo
...
+
+	private Map<String, Map> createUnboundAxesMap(VisualizerInfo visInfo, RuleInfo ruleInfo)
+	{
+		List unboundAxesList = ruleInfo.getUnboundAxesList()
+		if (unboundAxesList)
+		{
+			//Gather entries in unboundAxesList into a map containing two maps:
+			//one with cube names by scope key, one with scope values by scope key.
+			boolean hasScopeKeysToInclude
+			Map<String, Set<String>> cubesWithUnboundAxis = [:]
```
jdereg: This should be CaseInsenstiveMaps, not LinkedHashMaps [:]

src/main/groovy/com/cedarsoftware/util/RpmVisualizerRelInfo.groovy
```diff
@@ -376,61 +376,94 @@ class RpmVisualizerRelInfo extends VisualizerRelInfo
...
+			Map<String, Set<Object>> columnValuesForUnboundAxis = [:]
```
jdereg: This one is OK as is.  Column values are case-sensitive.

src/main/groovy/com/cedarsoftware/util/RpmVisualizerRelInfo.groovy
```diff
@@ -376,61 +376,94 @@ class RpmVisualizerRelInfo extends VisualizerRelInfo
...
+
+			unboundAxesList.each { MapEntry unboundAxis ->
+				String cubeName = unboundAxis.key as String
+				MapEntry axisEntry = unboundAxis.value as MapEntry
+				String axisName = axisEntry.key as String
+				if (includeScopeKey(visInfo, axisName))
 				{
-					sb.append("${OPTIONAL_SCOPE_AVAILABLE_TO_LOAD}${cubeName} for ${effectiveNameByCubeName}${sourceMessage}.")
+					Set<String> cubeNames = cubesWithUnboundAxis[axisName] ?: [] as Set
```
jdereg: Rather than do [] as Set, it should be new **CaseInsensitiveSet**.  Even if it were a LinkedHashSet, it would be quicker in byte code to do have new LinkedHashSet than [] as Set, which has to create an ArrayList, and then turn around and create a LinkedHashSet.  Lots of heap usage going on here.

src/main/groovy/com/cedarsoftware/util/RpmVisualizerRelInfo.groovy
```diff
@@ -376,61 +376,94 @@ class RpmVisualizerRelInfo extends VisualizerRelInfo
...
+					cubeNames << cubeName
+					cubesWithUnboundAxis[axisName] = cubeNames
+					Set<Object> columnValues = columnValuesForUnboundAxis[axisName] ?: [] as Set
```
jdereg: new LinkedHashSet() or new HashSet() would run faster than [] as Set.

src/main/groovy/com/cedarsoftware/util/RpmVisualizerRelInfo.groovy
```diff
@@ -442,19 +475,19 @@ class RpmVisualizerRelInfo extends VisualizerRelInfo
 	{
 		String cubeName = e.cubeName
 		String effectiveNameByCubeName = effectiveNameByCubeName
-		StringBuilder sb = new StringBuilder()
+		StringBuilder sb = new StringBuilder(ADDITIONAL_SCOPE_REQUIRED_TO_LOAD)
 		if (cubeName.startsWith(RPM_CLASS_DOT))
 		{
 			String cubeDisplayName = getCubeDisplayName(cubeName)
-			sb.append("${ADDITIONAL_SCOPE_REQUIRED_TO_LOAD}${effectiveNameByCubeName} of type ${cubeDisplayName}${sourceMessage}.")
+			sb.append("${effectiveNameByCubeName} of type ${cubeDisplayName}${sourceMessage}.")
```
jdereg: You can put getCubeDisplayName(cubeName) inside the ${} and eliminate the extra local variable and line

bheekin:
Ok, thanks for checking it through again. Will take a look at those comments tomorrow.

src/main/groovy/com/cedarsoftware/util/RpmVisualizerRelInfo.groovy
bheekin: Not sure what you mean... but the right side is the groovy short form for getEffectiveNameByCubeName() and is used for the coordinate not found message.

PR #396, opened 2017-01-31
Visualizer: More cleanup

bheekin:
1. Cleaned up the various ways an rpm class cube name is displayed (label, title, display title 1, display title 2, etc.) and fixed a few old TODOs in the process.
2. Changed the wording and format of the prompts for optional scope and required scope.
3. Fixed bug so that “Add to” link doesn’t show if class cannot be loaded.


src/main/groovy/com/cedarsoftware/util/VisualizerRelInfo.groovy
```diff
@@ -91,7 +92,7 @@ class VisualizerRelInfo
 
 	Set<String> getRequiredScope()
 	{
-		return targetCube.getRequiredScope(targetScope, [:] as Map)
+		return targetCube.getRequiredScope(scope, [:] as Map)
```
jdereg: [:] as Map - should this be a CasenInsentiveMap - in otherwords, if the keys are scope keys, it should be as scope keys should be handled case-insensitively at all times.

PR #400, opened 2017-02-01
Visualizer: Refine scope prompts

bheekin:
1. Reworked wording and format of scope prompts in toast yet again. 
   a. Now providing the unbound value the user may have provided but gotten no hit on. 
   b. If you hover over the optional scope dropdowns, you’ll see which cubes have the unbound axis, good for product model trouble shooting. 
   c. Created a new VisualizerScopeInfo to hold scope info.

2. If there’s required scope but no values to choose from (perhaps a declared scope key), the toast prompt now provides an input field to add the value. There is no example of this in UD.REF.APP, but refer to rpm.class.Policy in test.visualizer for an example.


src/main/groovy/com/cedarsoftware/util/RpmVisualizerRelInfo.groovy
```diff
@@ -278,29 +277,28 @@ class RpmVisualizerRelInfo extends VisualizerRelInfo
...
 		String targetCubeName = targetCube.name
 		if (targetCubeName.startsWith(RPM_CLASS_DOT))
 		{
-			detailsTitle = getCubeDisplayName(targetCubeName)
+			return getCubeDisplayName(targetCubeName)
 		}
 		else if (targetCubeName.startsWith(RPM_ENUM_DOT))
 		{
-			detailsTitle = "Valid values for field ${sourceFieldName} on ${getLabel(sourceCube.name)}".toString()
+			return "Valid values for field ${sourceFieldName} on ${getLabel(sourceCube.name)}".toString()
```
jdereg: You do need .toString() at the end of the GString.  Since this GString is returned thru a String return type, it will be converted to String.  If the return type was Object, then you would need the .toString()

src/main/webapp/js/visualize.js
```diff
@@ -204,15 +204,26 @@ var Visualizer = (function ($) {
...
         var target = e.target;
-        if (target.className.indexOf('missingScope') > -1) {
-            missingScope(target);
+        if (e.type === 'change' && target.className.indexOf('missingScopeSelect') > -1) {
+            missingScopeSelect(target);
         }
-        else if (target.className.indexOf('findNode') > -1) {
+        else if (e.type === 'change' && target.className.indexOf('missingScopeInput') > -1) {
+            missingScopeInput(target);
+        }
```
tympollack: These two conditionals can be nested
```
if type = change
  if classname missingScopeSelect
    doThing1
  else if classname missingScopeInput
    doThing2

PR #401, opened 2017-02-02
Clear note getting called in too many places, hiding info.

tympollack:


jdereg:
This change makes me a bit nervous.  The users are going to have to now close a lot more toast messages than before.  Some toasts should never be closed (The Visualizer Scope chooser or the results of a commit/update/rollback), some toasts should only show for a few seconds, and some toasts we should probably have die after 30-60 seconds (I think we may need to do this to some of 'one liner' messages we have).  I will merge this in, and then we can 'trim' back the annoying ones I suppose.

bheekin:
I’ll pull the code down on my local and will test out how it works with the visualizer. The visualizer does do some of its own “controlling” of toasts. For example, shows and then kills those “loading” and “stabilizing” messages, but keeps around scope prompt messages until closed by user or a reload occurs, etc.

tympollack:
Firstly, I didn't touch the visualizer toasts, mostly index and ncube view.

Secondly, users should not have to close more toasts, except for the few that don't have an expiration.

Finally, I only removed the clearNote for functions that didn't really make sense to clear it. I think there are still 7 places that use it such as commit, rollback, server stats, http headers, and a couple others. But leaving it in so many places meant that many toasts (which are mostly error notifications) were being wiped out before they could even be seen.

PR #409, opened 2017-02-08
Overflow tab would not open cube properly.

tympollack:


jdereg:
I suggest updating the Help text at the top of the modal (if it is not already there) to mention support of copy/paste here.  Most users probably do not know they can copy/paste here.

PR #413, opened 2017-02-16
Allow impersonation admin tool in NCE.

tympollack:


src/main/webapp/js/index.js
```diff
@@ -2183,15 +2185,77 @@ var NCE = (function ($) {
         }
     }
 
+    function showHideImpersonation(isAdmin) {
```
gaigstroliac: This is a very long function with lots of nested code blocks (pyramid of doom). Could it be refactored to simplify it a bit?

tympollack: Probably, but I have a todo here and for the tab menu to refactor into more dynamic/reusable (and straight html instead of jquery for constructing elements), so it'll be done then.

PR #416, opened 2017-02-17
Enable "Clear Server Cache" for all users except when branch is HEAD

bheekin:


src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
```diff
@@ -1256,14 +1256,17 @@ class NCubeController extends BaseController
...
-        if (!isAppAdmin(appId))
+        if (isAppAdmin(appId))
         {
-            return
+            nCubeService.clearCache(appId)
+            clearAppCache(appId.tenant)
+            clearVersionCache(appId.app)
+            clearBranchCache(appId)
+        }
+        else if (ApplicationID.HEAD != appId.branch)
```
tympollack: You can use `else if (!appId.head)` here.

PR #419, opened 2017-02-21
Revert "Enable "Clear Server Cache" for all users except when branch is HEAD"

jdereg:
Reverts jdereg/n-cube-editor#416


src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
```diff
@@ -1256,17 +1256,14 @@ class NCubeController extends BaseController
     void clearCache(ApplicationID appId)
     {
         appId = addTenant(appId)
-        if (isAppAdmin(appId))
+        if (!isAppAdmin(appId))
```
gaigstroliac: It seems like it would have been easier and more efficient to have simply removed the second `else if`.

This would have kept the first `if` positive and would not have required adding a return statement.

PR #420, opened 2017-02-21
Revert "Middle click should close cube tab."

jdereg:
Reverts jdereg/n-cube-editor#415

### DIFF
--- a/src/main/webapp/js/index.js
+++ b/src/main/webapp/js/index.js
@@ -710,35 +710,30 @@ var NCE = (function ($) {
             title: cubeInfo.slice(0, CUBE_INFO.TAB_VIEW).join(' - '),
             template: '<div class="tooltip" role="tooltip"><div class="tooltip-arrow"></div><div class="tooltip-inner tab-tooltip"></div></div>'
         });
-        li.on('click auxclick', function(e) {
-            var target, isClose, isDropdown, self;
-            self = $(this);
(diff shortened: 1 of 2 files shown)


src/main/webapp/js/index.js
```diff
@@ -710,35 +710,30 @@ var NCE = (function ($) {
...
-                    if (!self.find('ul').length) {
-                        addTabDropdownList(self, cubeInfo);
+                    if (!xthis.find('ul').length) {
+                        addTabDropdownList(xthis, cubeInfo);
                     }
-                    self.find('.ncube-tab-top-level')
+                    xthis.find('.ncube-tab-top-level')
                         .addClass('dropdown-toggle')
                         .attr('data-toggle', 'dropdown');
                     $(document).one('click', function() { // prevent tooltip and dropdown from remaining on screen
```
gaigstroliac: I can see that this code was already here, but was it really supposed to be `one` and not `on`?

tympollack: No. This calls over to close the tab menu if the user clicks off of it. That only needs to happen once after the menu is open, once it's closed, we don't want to keep calling it every time the user clicks around.

gaigstroliac: Oh, sorry, I didn't even realize that `.one` was an actual method on a jquery object. I could have sworn that I'd accidentally used the `.one` method in the past (typo) and it gave me an error or warning, but perhaps that was on jqlite from angular or I was simply using it wrong or something. 

Anyway, glad it wasn't a typo.

PR #427, opened 2017-02-28
Preliminary commit request changes; bug fixes.

tympollack:
@jdereg Commit requests work, as well as the links. However, for right now, the requests are only viewable on the modal, and clicking them reveals the cube list they are committing. More functionality to follow, but this gets us in the door and able to start using them.

### DIFF
--- a/src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
+++ b/src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
@@ -41,6 +41,7 @@ import com.cedarsoftware.util.StringUtilities
 import com.cedarsoftware.util.SystemUtilities
 import com.cedarsoftware.util.ThreadAwarePrintStream
 import com.cedarsoftware.util.ThreadAwarePrintStreamErr
+import com.cedarsoftware.util.UniqueIdGenerator
 import com.cedarsoftware.util.Visualizer
 import com.cedarsoftware.util.io.JsonReader
 import com.cedarsoftware.util.io.JsonWriter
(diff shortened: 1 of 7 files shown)


src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
```diff
@@ -1352,6 +1339,93 @@ class NCubeController extends BaseController
...
+        Object[] commitDtos = allDtos.findAll {Object dto -> commitNames.contains((dto as NCubeInfoDto).name)}
+
+        commitsCube.setCell('processing', [property:'status'])
+        commitsCube.setCell(user, [property:'commitUser'])
+        commitsCube.setCell(new Date().format('M/d/yyyy HH:mm:ss'), [property:'commitTime'])
+        NCubeManager.getPersister().updateCube(sysAppId, commitsCube, user)
+
+        Object commitBranchResponse = commitBranch(commitAppId, commitDtos)
+        String message = null
+        if (JsonCommandServlet.servletRequest.get().getAttribute(JsonCommandServlet.ATTRIBUTE_STATUS))
```
jdereg: I am not sure I understand this piece - where it checks the 'status'.  This value is set by the advice when the controller returns (to the command servlet) - so the command servlet knows how to create the return envelope.

jdereg:
Before I merge this - can the general user still commit n-cubes without using the 'link'?  I don't want to force this until we have explained this to everyone (set their expectations).

tympollack:
Latest commit removes the hiding of the commit button, so that will still be available until we need to turn it off.

src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
tympollack: If commitBranch() fails, the status will already be set to false with a failure message. I wanted to intercept that to pass it back to the user but also do stuff based on whether it succeeds or fails (need to update the commit request info).

PR #429, opened 2017-03-02
Visualizer: Scope prompt refactor

bheekin:
Here’s the PR for the scope prompt refactor. Lots of changes… Sorry, the PR is so large, but this change is an “all or nothing” type change (or “a lot or nothing”). Many of the changes files are test json.

I have a ton of commits as part of this PR too. **I suggest squashing the commits upon merge.** 

**Changes (high level):**

1.	The editable input field for scope has been removed. Replaced with a toast that prompts the user to select or enter scope as needed.
2.	The scope toast contains information about the scope for the top node being visualized and any other scope needed to load its target classes _without traits_. Hover over input fields in the scope toast to see more info about the scope.
3.	The details panel (shows when clicking a node) displays scope prompts for individual nodes. There are two types of node prompts:
    a.	Prompts for scope that is required to load the class _without traits_. This same scope displays in the scope toast and is regarded as part of the “graph scope”.  When the user changes this kind of prompt, the visualizer reloads completely and the scope toast is updated.
    b.	Prompts for scope that is required to load the class _with traits_. These prompts only show after the user has clicked “Show traits”. This scope does not display in the scope toast (doesn’t affect the graph as a whole). When the user changes this kind of prompt, only the details panel reloads with updated scope info for the node in question.
4.	The user provided graph scope is still saved to local storage as before. But now, in order to rebuild the scope prompts with up to date available scope values, the visualizer loads classes without the user-provided scope initially. As it encounters a missing or unbound scope key, it grabs the available scope values, then reloads the class in question with the user provided scope.
5.	The scope toast stays put as scope is reloaded (instead of being cleared and recreated) and is updated dynamically with new scope info.
6.	The separate toasts that previously flashed up for “loading…”, “stabilizing”, etc. have been eliminated. These messages now show in the main scope toast. 
7.	The funky scope button on the toast closes the toast. The same scope button is on the visualizer page itself and allows the user to open or close the scope toast. 
8.	Moved the details panel from “east” to “west” so it’s not covered by the scope toast.

**TODOs:** 

1.	Add functionality for “in scope” scope values for r:exists. Requires n-cube change as well to add coordinate info the unbound scope info on RuleInfo and to InvalidCoordinateException. 
2.	Disabled the ability to show cell values for regular n-cubes (non-rpm). Need to revisit and apply same scope prompt concepts as for rpm cubes.
3.	Additional code cleanup… may be able to combine some methods that do similar work. Also considering moving target scope and availableTargetScope on VisualizerRelInfo into VisualizerScopeInfo.


bheekin:
After team discussions, we've decided to take a different approach with the scope prompts. Closing the PR to rework.

PR #430, opened 2017-03-02
Multiple changes.

tympollack:


src/main/webapp/js/index.js
```diff
@@ -3846,6 +3846,9 @@ var NCE = (function ($) {
     function showActiveBranch() {
         if (getBranchNames().indexOf(_selectedBranch) > -1) {
             addToVisitedBranchesList(appIdFrom(_selectedApp, _selectedVersion, _selectedStatus, _selectedBranch));
+        } else if (_selectedBranch == head) {
```
gaigstroliac: double equals

tympollack: Changed with commit. Why we have code reviews. :)

PR #431, opened 2017-03-03
Multiple changes.

tympollack:


src/main/groovy/com/cedarsoftware/controller/NCubeController.groovy
```diff
@@ -1435,7 +1472,9 @@ class NCubeController extends BaseController
         for (NCubeInfoDto dto : dtos)
         {
             NCube cube = nCubeService.loadCubeById(Long.parseLong(dto.id))
-            results.add(cube.getMap([property:[] as Set]))
+            Map commitInfo = cube.getMap([property:[] as Set])
+            commitInfo.put('txid', dto.name.substring(3))
```
jdereg: commInfo['txid'] = dto.name.substring(3) instead of .put()

src/main/webapp/index.html
```diff
@@ -334,7 +334,7 @@ <h3 id="revisionHistoryLabel" class="modal-title"></h3>
                 <button type="button" class="close" data-dismiss="modal" aria-hidden="true">&times;</button>
                 <h3 id="view-commits-label" class="modal-title">Commit Requests</h3>
```
jdereg: We need to talk about the naming of things here.  We need a more accurate name than 'Commit request'. These are really 'Pull requests'.  In NCE, it automatically commits to your branch on every change.  With new changes, the user creates a 'PR' not 'CR'.

PR #435, opened 2017-03-10
PR view updates on heartbeat if open.

tympollack:


src/main/webapp/js/index.js
```diff
@@ -4400,11 +4400,9 @@ var NCE = (function ($) {
             txid = result.data;
             urlPrefix = document.URL;
             urlPrefix = urlPrefix.substring(0, urlPrefix.lastIndexOf('/'));
-            gitUrl = urlPrefix + '/cmd/ncubeController/honorCommit/?json=["' + txid + '"]';
             viewUrl = urlPrefix + '/#/viewCommit/' + txid;
-            html = 'Pull Request View Link:<br><a href="#" onclick="NCE.closeOpenModal();NCE.viewCommits(true,\'' + txid + '\');">' + viewUrl + '</a><hr>'
-                 + 'GitHub Link (WARNING: This will merge the pull request!):<br>' + gitUrl;
-            showNote(html, 'Commit Link', null, NOTE_CLASS.HAS_EVENT);
+            html = 'Pull Request View Link:<br><a href="#" onclick="NCE.closeOpenModal();NCE.viewCommits(true,\'' + txid + '\');">' + viewUrl + '</a>';
```
gaigstroliac: I know that this is just fixing some existing things, but this feels like it should not have so much logic here. 

What do you think about creating a function for this?

e.g. 
``` javascript
function pullRequestView(txid){
  NCE.closeOpenModal();
  NCE.viewCommits(true, txid);
}
...
html = 'Pull Request  View Link:<br><a href="#" onclick="pullRequestView(\'' + txid + '\')">' + viewUrl + '</a>';
```

tympollack: Fixed.

PR #436, opened 2017-03-14
POC for FormBuilder replacing Update Axis.

tympollack:


gaigstroliac:
I didn't look at every single line, but overall, very elegant looking code. 👍

tympollack:
Thanks, @gaigstroliac . The goal of the form builder ultimately is to allow plugins to easily collect form data from the user (inputs for scripts, adding objects to the product model, etc.). As a side effect, we can leverage the same code to dynamically populate our basic existing modals and clean up our static js/html.

PR #437, opened 2017-03-16
Visualizer: Refactor how scope is handled and displayed

bheekin:


src/main/groovy/com/cedarsoftware/util/RpmVisualizerRelInfo.groovy
```diff
@@ -351,138 +368,211 @@ class RpmVisualizerRelInfo extends VisualizerRelInfo
 				handleException(t, visInfo)
 			}
 		}
+		addRequiredScopeKeys(visInfo)
+		retainUsedScope(visInfo, output)
+		if (loadAgain)
+		{
+			return loadCube(visInfo, output)
+		}
```
jdereg: If the code above is supposed to only recurse-twice, this method should be broken up.  It is mixing the loading process with the re-try strategy.  Can the two be separated, and a simple if and a second try made?  It would be a lot cleared, and more importantly, non-recursive.

src/main/groovy/com/cedarsoftware/util/RpmVisualizerRelInfo.groovy
```diff
@@ -351,138 +368,211 @@ class RpmVisualizerRelInfo extends VisualizerRelInfo
...
+			{
+				return
 			}
+			nodeDetailsMessages << "Defaults were used for some scope keys.${DOUBLE_BREAK}".toString()
 		}
-		return scopeInfo.axisNames
 	}
 
-	private static addValuesForUnboundAxis(Collection valuesToAdd, String axisName, Map<String, Set> valuesMap)
+	private void handleCoordinateNotFoundException(CoordinateNotFoundException e, VisualizerInfo visInfo)
```
jdereg: These three methods look almost identical.  Can a 4th method be added that 'does the work' and these 3 call it with args?

src/main/groovy/com/cedarsoftware/util/RpmVisualizerRelInfo.groovy
```diff
@@ -351,138 +368,211 @@ class RpmVisualizerRelInfo extends VisualizerRelInfo
...
+
+	@Override
+	protected void populateScopeDefaults(VisualizerInfo visInfo)
+	{
+		Map<String, Object> scopeDefaults = new CaseInsensitiveMap()
+		String scopeValue = visInfo.inputScope[EFFECTIVE_VERSION] ?: appId.version
+		addScopeDefault(scopeDefaults, EFFECTIVE_VERSION, scopeValue)
+		//loadAvailableScopeValuesEffectiveVersion(visInfo) TODO: Very slow, commenting out for now
+
+		if (targetCube.getAxis(AXIS_TRAIT).findColumn(R_SCOPED_NAME))
```
jdereg: The targetCube.getAxis(AXIS_TRAIT).findColumn(R_SCOPE_NAME) is done many times.  Can this be put into a method like isScopedClass(targetCube) and then all the places call this?

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -19,127 +17,123 @@ class Visualizer
...
+	 * @return   VisualizerInfo visInfo, information about the visualization
 	 */
-	Map getCellValues(ApplicationID applicationID, Map options)
+	Map<String, Object> loadScopeChange(ApplicationID applicationID, Map options)
 	{
 		appId = applicationID
-		VisualizerRelInfo relInfo = new VisualizerRelInfo(appId, options.node as Map)
-		return getCellValues(relInfo, options)
+		VisualizerInfo visInfo = getVisualizerInfo(options)
+		Map selectedNode = new LinkedHashMap(visInfo.nodes[visInfo.selectedNodeId])
```
jdereg: LinkedHashMap here - is that correct or should it be CaseInsenstiveMap?  Just wanted to make sure we are not being case-sensitive where we should not be.

src/main/groovy/com/cedarsoftware/util/Visualizer.groovy
```diff
@@ -19,127 +17,123 @@ class Visualizer
...
 	{
 		VisualizerInfo visInfo = options.visInfo as VisualizerInfo
 		if (!visInfo || visInfo.class.name != this.class.name)
 		{
-			visInfo = new VisualizerInfo(appId, options)
+			visInfo = new VisualizerInfo(appId)
+		}
+		else
+		{
+			visInfo.appId = appId
```
jdereg: What is this method doing?  Is it trying to update the appId or create a new VisualizerInfo instance?  Seems like a vague mix of behavior.

src/main/groovy/com/cedarsoftware/util/VisualizerCellInfo.groovy
```diff
@@ -100,28 +102,27 @@ class VisualizerCellInfo
...
 		}
 		else
 		{
-			title = 'An error occurred during the execution of the cell.'
+			title = 'An error occurred during the execution of the cell'
 			listItemClassName = DETAILS_CLASS_EXCEPTION
-			String targetMsg = "coordinate ${coordinateString}"
-			mb.append(helper.handleException(t, targetMsg))
+			mb.append("An exception was thrown while loading the coordinate.${DOUBLE_BREAK}")
+			mb.append(helper.handleException(t))
```
jdereg: A log.warn or at least log.info() should log an exception that you are not expecting, otherwise these are just swallowed and it makes bugs harder to find.

src/main/groovy/com/cedarsoftware/util/VisualizerInfo.groovy
```diff
@@ -16,89 +13,104 @@ import static com.cedarsoftware.util.VisualizerConstants.*
...
+            removeSourceEdges()
+            removeTargets(edges)
+            removeTargets(nodes)
+            maxLevel = 1
+            availableGroupsAllLevels = new LinkedHashSet()
+        }
     }
 
-    boolean addMissingMinimumScope(String scopeKey, String value, String message, Set<String> messages)
+    private void removeTargets(Map<Long, Map> nodes)
```
jdereg: For both removeTargets and removeSourceEdges() I believe you can just use groovy's removeAll.  http://mrhaki.blogspot.com/2015/09/groovy-goodness-removing-elements-from.html

src/main/groovy/com/cedarsoftware/util/VisualizerInfo.groovy
```diff
@@ -137,48 +148,55 @@ class VisualizerInfo
...
+            nodeCountAtLevel++
+            levelNodeCount[nodeLevel as int] = nodeCountAtLevel
+        }
 
-    Set<Object> getOptionalScopeValues( String cubeName, String scopeKey)
-    {
-        return getScopeValues(optionalScopeValues, cubeName, scopeKey)
+        //Determine cumulative node count at each level + determine defaultLevel
+        levelCumulativeNodeCount = [:]
+        for (Integer i = 1; i < levelNodeCount.size() + 1; i++)
```
jdereg: Please use int instead of Integer in for-loop index.

src/main/groovy/com/cedarsoftware/util/VisualizerRelInfo.groovy
```diff
@@ -200,35 +241,35 @@ class VisualizerRelInfo
...
+		if (!visInfo.requiredScopeKeysByCube.containsKey(cubeName))
 		{
-			visInfo.requiredScopeKeys[cubeName] = requiredScope
-			visInfo.optionalScopeKeys[cubeName] = targetCube.getOptionalScope(scope, [:])
+			visInfo.requiredScopeKeysByCube[cubeName] = requiredScope
 		}
 	}
 
-	Map<String, Object> createEdge(int edgeCount)
+	protected Map<String, Object> createEdge(Long edgeId)
```
jdereg: What values does edgeId take on?  A Long (or long) cannot be passed to the front-end successfully if it is great than 15 digits, and a long can be 19 digits.  This is a very serious issue and needs a discussion.  We have been burned many times by longs being passed to the front end and the front-end losing part of the value.

src/main/groovy/com/cedarsoftware/util/VisualizerRelInfo.groovy
```diff
@@ -240,49 +281,259 @@ class VisualizerRelInfo
...
+		return getAllColumnValues(cube, axisName)
+	}
+
+	protected static Set<Object> getAllColumnValues(NCube cube, String axisName)
+	{
+		Set values = new LinkedHashSet()
+		Axis axis = cube.getAxis(axisName)
+		if (axis)
+		{
+			for (Column column : axis.columns)
```
jdereg: do you want axis.columns or axis.columnsWithoutDefault?

src/main/groovy/com/cedarsoftware/util/VisualizerRelInfo.groovy
```diff
@@ -240,49 +281,259 @@ class VisualizerRelInfo
...
+		sb.append(nodeScopeMessage)
+		return sb.toString()
+	}
+
+	private StringBuilder getNodeDetailsMessageSet()
+	{
+		StringBuilder sb = new StringBuilder()
+		if (nodeDetailsMessages)
+		{
+			nodeDetailsMessages.each { String message ->
```
jdereg: can remove if and just use nodeDetailsMessages?.each { ... }

PR #460, opened 2017-04-21
Add status / branch to reference/transform to allow snapshots.

tympollack:


src/main/resources/static/js/nce-builder-options.js
```diff
@@ -282,6 +282,7 @@ var NCEBuilderOptions = (function () {
                             listeners: {
                                 change: function() {
                                     FormBuilder.findElementByKey('refVer').trigger('populate');
+                                    FormBuilder.findElementByKey('refBranch').trigger('populate');
```
gaigstroliac: I'm seeing a lot of repetition here. 

Any chance this could all be wrapped in a single function?

ex.
```javascript
triggerPopulate('refVer');
triggerPopulate('refBranch');
...
// OR
['refVer','refBranch','refCube','refAxis'].forEach(triggerPopulate); // less efficient, I know
...
function triggerPopulate(key){
    FormBuilder.findElementByKey(key).trigger('populate');
}
```

PR #484, opened 2017-05-30
Batch update reference axis UI changes.

tympollack:


src/main/resources/static/js/nce-builder-options.js
```diff
@@ -1170,6 +1137,232 @@ var NCEBuilderOptions = (function () {
...
+     * additional required options:
+     *  refAxList
+     *  isTransform
+     *  appSelectList
+     *  populateVersionFunc
+     *  populateBranchFunc
+     *  populateCubeFunc
+     */
+    function referenceAxisUpdater(opts) {
+        var ret = {
```
gaigstroliac: It feels like this could benefit from being broken down into smaller functions that compose the whole. For instance, the formInputs property could be assigned like formInputs: generateFormInputs(opts) and similar things could be done for the update sections.

PR #485, opened 2017-06-01
Add column sorting functionality to FormBuilder tables.

tympollack:


src/main/resources/static/js/nce-builder-options.js
```diff
@@ -1280,44 +1280,51 @@ var NCEBuilderOptions = (function () {
 
             columns.srcCubeName = {
                 heading: 'Source Cube',
-                    type: FormBuilder.INPUT_TYPE.READONLY,
-                    css: {width: '32%', overflow:'hidden', 'text-overflow':'ellipsis', 'background-color':'white'}
+                type: FormBuilder.INPUT_TYPE.READONLY,
+                sortable: true,
+                css: {width: '32%', overflow:'hidden', 'text-overflow':'ellipsis', 'background-color':'white'}
```
gaigstroliac: All the repetition here seems like it could be wrapped up into a function.

gaigstroliac:
```javascript
function makeColumn(heading, width){
  return {
    heading: heading,
    type: FormBuilder.INPUT_TYPE.READONLY,
    sortable: true,
    css: {width: width, overflow:'hidden', 'text-overflow':'ellipsis'}
  };
}
```

tympollack:
Apparently I can't reply to the variable definition comment. Functions are hoisted to top of scope before vars, so defining either var or function first yields the same result. Regardless, both are defined before any executable lines.

gaigstroliac:
@tympollack 
I see what you mean.

I've tended to treat variables and functions differently because even though variables are hoisted after functions, they are initialized on the line in which they reside. Whereas, functions are hoisted _**and**_ defined at the top of the function at the same time(unless they are defined using `var fn = function(){}`, syntax) .

For that reason, I'm usually in the habit of declaring my functions below any code I have that runs "inline" so that code doesn't get lost in the mix (plus, I hate scrolling down past hundreds or thousands of lines of code to see what the initialization code looks like). But I guess, since that's not the paradigm being used here, that I wouldn't change it in this case.

About the only time I do put inline code below the top is when it is clearly intended to augment the defined function in some way (like the code you have on line 1341 here), but I still tend to put the actual variable declarations at the top if I have them. Just a preference I guess. 

ex:
```javascript
function(){
  var myVar, someVar;
  myVar = 'A';
  someVar = 10;
  return doSomethingWith(myVar, someVar);

  function doSomethingWith(a, b){
    return String(a) + String(b);
  }
}
```

tympollack:
@gaigstroliac Latest commit better?

gaigstroliac:
@tympollack 
That's beautiful man. 👍

PR #486, opened 2017-06-02
Add copy support for cube/axis/column names.

tympollack:


src/main/resources/static/js/ntwobe.js
```diff
@@ -3095,47 +3095,48 @@ var NCubeEditor2 = (function ($) {
...
-                    clipData += '"' + content.replace(/"/g, '""') + '"';
-                } else {
-                    clipData += content;
-                }
+                    if (content.indexOf('\n') > -1) {
+                        // Must quote if newline (and double any quotes inside)
+                        clipData += '"' + content.replace(/"/g, '""') + '"';
+                    } else {
+                        clipData += content;
+                    }
```
gaigstroliac: ternary?

```javascript
clipData += (content.indexOf('\n') > -1) ? ('"' + content.replace(/"/g, '""') + '"') : content;
```

src/main/resources/static/js/ntwobe.js
```diff
@@ -3144,24 +3145,75 @@ var NCubeEditor2 = (function ($) {
...
+
+        function getAxisNames(axisNumStart, axisNumEnd) {
+            var i, len;
+            var vals = '';
+            for (i = axisNumStart, len = axisNumEnd; i <= len; i++) {
+                if (vals) {
+                    vals += '\n';
+                }
+                vals += axes[i].name;
+            }
```
gaigstroliac: I can't recall if I've ever checked to efficiency of it, but for stuff like this, I often find it easier to add items to an array and then return with a `join`. That way you don't have to constantly do a check for the last item.

```javascript
var vals = [];
for (i = axisNumStart, len = axisNumEnd; i <= len; i++) {
  vals.push(axes[i].name);
}
return vals.join('\n');
```

for even more terse and readable, but at the likely expense of efficiency
```javascript
  return axes
    .slice(axisNumStart, asixNumEnd+1)
    // I would so like to write the next line as .map(axis => axis.name)
    .map(function(axis){ 
      return axis.name; 
    })
    .join('\n');
```

gaigstroliac:
Looks like some of my comments were only about things you indented, so take them for what they're worth.

src/main/resources/static/js/ntwobe.js
tympollack: That's true about array join, I do this a lot and never thought about using that...

As for slice, map, join chain, I'll have to check efficiency, but does seem more readable.

src/main/resources/static/js/ntwobe.js
tympollack: Maybe, looks kinda messy tho.

tympollack:
@gaigstroliac yeah the bulk of it was wrapping existing code into an if, so it's not all new.

PR #493, opened 2017-06-22
Remove ncube old and renaming ntwobe.

tympollack:
We will need to update sys.menu in UD.REF.APP when this gets built.


jdereg:
This change now conflicts because I have delayed merging it.  I am going to close it for now.  There is no rush on this change.

PR #501, opened 2017-07-10
Optimize pull request modal js.

tympollack:


src/main/resources/static/js/index.js
```diff
@@ -3450,14 +3450,16 @@ var NCE = (function ($) {
...
         tds = _viewPullRequestsList.find('tr:visible').find('td:nth-child(' + idx + ')');
         for (i = 0, len = tds.length; i < len; i++) {
             td = tds[i];
             row = $(td).parent();
             if (td.innerHTML.indexOf(filterVal) === -1) {
-                row.hide();
+                hideRows.add(row);
             }
         }
+        hideRows.hide()
```
gaigstroliac: I'm a little concerned about what I see here. 

it looks like `hideRows` is an Array type, but Arrays don't have `add` and `hide` methods on the `prototype`, and it's considered bad juju to add prototype methods to built-in types unless they are a polyfill for a spec that is, or at least is likely to be, finalized for that built-in type.

PR #505, opened 2017-07-14
Language detection for Ace popout.

tympollack:


src/main/resources/static/js/ntwobe.js
```diff
@@ -3525,16 +3525,38 @@ var NCubeEditor2 = (function ($) {
...
+    function detectLanguage(text) {
+        var cssRegEx = new RegExp('[\\#\\.\\w\\-\\,\\s\\n\\r\\t:]+(?=\\s*\\{)', 'gi');
+        var htmlRegEx = new RegExp('<(?:br|p)[^>{]*>|</\\w+\\s*>', 'gi');
+        if (!text.indexOf('{')) {
+            return 'json';
+        }
+        if (text.indexOf('function') > -1 || text.indexOf('var') > -1) {
+            return 'javascript';
+        }
+        if (htmlRegEx.exec(text)) {
```
gaigstroliac: `test` would probably be more efficient than `exec` for what you are trying to do here.

gaigstroliac: I would verify that it works as expected though. I've been doing some testing and there seem to be some instances where things that exec would be true for are not true for test.

Also, might want to declare/compile the regex's outside of this function somewhere.

jdereg:
This is nice win for the end-user - great update.

PR #511, opened 2017-07-28
moving nexusReleaseRepoId to a property

irhiggs:


pom.xml
```diff
@@ -103,7 +104,7 @@
 
     <distributionManagement>
```
tympollack: Should nexus info be a part of n-cube-editor, or should there be a forked version that includes these properties?

PR #512, opened 2017-07-31
Fix for changes made to checkPermissions on controller.

tympollack:


src/main/resources/static/js/ntwobe.js
```diff
@@ -4472,7 +4472,9 @@ var NCubeEditor2 = (function ($) {
             clearFilters();
             markCubeModified();
             nce.loadCube();
-            editColumns(axisName);
+            if (!data.isRef) {
```
tympollack: This was changed to prevent edit columns modal if the new axis is a ref axis.

PR #516, opened 2017-08-07
Fix for Modal formatting with data tables

Xanmia:
- Removed html table creation for data header, moved header row to append dataTable.
- Affects Edit Axis Metadata modal and Filter Data modal.
- Updated Filter data modal options to MODAL_SIZE.LARGE to accommodate new formatting.

### DIFF
--- a/src/main/resources/static/js/form-builder.js
+++ b/src/main/resources/static/js/form-builder.js
@@ -603,14 +603,13 @@ var FormBuilder = (function ($) {
     function buildTable(data, tableOpts) {
         var columns, columnKeys, headingRow, c, cLen, column, header, key;
         var style = tableOpts.css || { margin: '0 auto' };
-        var headerTable = $('<table/>').css(style);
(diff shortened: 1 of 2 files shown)


src/main/resources/static/js/form-builder.js
```diff
@@ -603,14 +603,13 @@ var FormBuilder = (function ($) {
     function buildTable(data, tableOpts) {
         var columns, columnKeys, headingRow, c, cLen, column, header, key;
         var style = tableOpts.css || { margin: '0 auto' };
-        var headerTable = $('<table/>').css(style);
```
tympollack: @Xanmia The reason for the header and data being separated tables are so headers are stationary while the table scrolls. With this change, table headers will be hidden if the table is scrolled.

Xanmia: ok, will looking into formatting with table scrolls.

PR #524, opened 2017-08-16
- Separated server stats from hearbeat() into getServerStats()

jsnyder4:


tympollack:
Note that this requires n-cube 4.0.21

src/main/resources/static/js/constants.js
```diff
@@ -307,6 +307,7 @@ var CONTROLLER_METHOD = {
     GET_REFERENCES_FROM: 'getReferencesFrom',
     GET_REQUIRED_SCOPE: 'getRequiredScope',
     GET_REVISION_HISTORY: 'getRevisionHistory',
+    GET_SERVER_STATS: 'getServerStats',
```
gaigstroliac: Is `HEARTBEAT` still being used somewhere? If not we might want to get rid of that constant

tympollack: Yes, it's still used on heartbeat. This PR removes extensive logic out of heartbeat that's not needed so it's not running every 60 sec / active user.

gaigstroliac: I just didn't notice the actual constant being used elsewhere.

tympollack: You're right, currently it's not bc the only other place calling heartbeat is in a separate web worker js file that can't reference the constants file. However, @jsnyder4 chose to leave it bc it's still a valid API on the controller.

PR #566, opened 2018-04-24
Commit option hidden when user had update, but not commit permissions…

tympollack:
…; need commit option to create pull request.


tympollack:
@jdereg this needs merged and redeployed to align with changes in n-cube

PR #569, opened 2018-06-21
add warning when releasing cubes with no next version

tympollack:


tympollack:
recommended by @jsnyder4

